Issue #141, opened 2026-03-04
feat: improve agent output copyability in TUI


devinrosen:
## Summary

When an agent outputs a command (curl, bash, etc.), the text is wrapped inside the Agent Activity box making it difficult to copy cleanly without manual cleanup of line breaks and box artifacts.

## Workaround (Today)

tmux copy mode works but is awkward: `C-b [` → navigate → `Space` to start selection → `Enter` to copy. The TUI borders disappear from the selection but the flow is unintuitive.

### 1. "Copy last code block" keybinding (low effort)

Add a keybinding (e.g. `y`) that:
1. Scans the current run's log file for the most recent fenced code block
2. Copies its contents directly to the clipboard via `pbcopy` (macOS) / `xclip` (Linux)
3. Shows a brief confirmation in the status bar: "Copied to clipboard"

The log file path is already stored on every `AgentRun` — this is mostly string parsing + a shell call. Solves the most common case (copying the last command the agent gave you) in one keypress.

### 2. Expand-to-detail view (medium effort)

Pressing `Enter` or `e` on an agent event opens a full-screen popup showing the raw event text:
- Word-wrap disabled
- Vertically and horizontally scrollable
- `Esc` or `q` to dismiss

Makes it easy to read and copy any agent output, not just code blocks. Follows the same expand pattern used elsewhere in ratatui apps.

## Implementation Notes

- Start with the keybinding — it's a quick win for the most common case
- The expand view is a natural follow-on and can be a separate PR
- Horizontal scroll in the activity box (disabling word-wrap) may also help and is worth considering alongside the expand view

PR #156, opened 2026-03-05
feat: improve agent output copyability in TUI (#141)

devinrosen:
Add two keybindings to enhance copying agent output:

1. **Copy last code block (y)**: Scans the agent log file for the most
   recent fenced code block (```...```) and copies its contents to clipboard
   via pbcopy (macOS) / xclip / xsel (Linux). Shows status confirmation.

2. **Expand-to-detail view (e)**: Opens a full-screen popup showing the
   raw event text with no word wrapping. Supports vertical (j/k) and
   horizontal (h/l) scrolling, plus vim navigation (gg/G for top/bottom).
   Press q/Esc to dismiss.

Implementation details:
- Added CopyLastCodeBlock and ExpandAgentEvent actions
- Added EventDetail modal variant for the expand view
- extract_last_code_block() helper parses fenced code from log content
- Updated keybindings in input.rs (y with log requirement, e always available)
- Updated help screen and action hints in worktree detail view
- Added 4 unit tests for code block extraction logic
- All tests pass, clippy -D warnings clean

### DIFF
--- a/conductor-tui/src/app.rs
+++ b/conductor-tui/src/app.rs
@@ -24,6 +24,11 @@ use crate::state::{
 };
 use crate::ui;
 
+/// Maximum scroll offset for a text body (total lines minus one visible line).
+fn max_scroll(line_count: usize) -> u16 {
+    line_count.saturating_sub(1) as u16
+}
+
 /// Derive a worktree slug from a ticket's source_id and title.
 /// Format: `{source_id}-{slugified-title}`, e.g. `15-tui-create-worktree`.
 /// Title portion is truncated to keep the total slug under ~40 chars.
@@ -314,34 +319,16 @@ impl App {
             Action::ViewAgentLog => self.handle_view_agent_log(),
             Action::CopyLastCodeBlock => self.handle_copy_last_code_block(),
             Action::ExpandAgentEvent => self.handle_expand_agent_event(),
-            Action::AgentActivityDown => match self.state.modal {
-                Modal::EventDetail {
-                    ref mut scroll_offset,
-                    line_count,
-                    ..
-                } => {
-                    let max = line_count.saturating_sub(1) as u16;
-                    *scroll_offset = (*scroll_offset + 1).min(max);
+            Action::AgentActivityDown => {
+                let len = self.state.data.agent_activity_len();
+                let cur = self.state.agent_list_state.borrow().selected().unwrap_or(0);
+                if len > 0 && cur + 1 < len {
+                    self.state.agent_list_state.borrow_mut().select_next();
                 }
-                _ => {
-                    let len = self.state.data.agent_activity_len();
-                    let cur = self.state.agent_list_state.borrow().selected().unwrap_or(0);
-                    if len > 0 && cur + 1 < len {
-                        self.state.agent_list_state.borrow_mut().select_next();
-                    }
-                }
-            },
-            Action::AgentActivityUp => match self.state.modal {
-                Modal::EventDetail {
-                    ref mut scroll_offset,
-                    ..
-                } => {
-                    *scroll_offset = scroll_offset.saturating_sub(1);
-                }
-                _ => {
-                    self.state.agent_list_state.borrow_mut().select_previous();
-                }
-            },
+            }
+            Action::AgentActivityUp => {
+                self.state.agent_list_state.borrow_mut().select_previous();
+            }
             // Scroll navigation (all views + discover modals)
             Action::GoToTop => match self.state.modal {
                 Modal::EventDetail {
@@ -366,7 +353,7 @@ impl App {
                     line_count,
                     ..
                 } => {
-                    *scroll_offset = line_count.saturating_sub(1) as u16;
+                    *scroll_offset = max_scroll(line_count);
                 }
                 Modal::GithubDiscoverOrgs {
                     ref orgs,
@@ -681,6 +668,13 @@ impl App {
 
     fn move_up(&mut self) {
         match self.state.modal {
+            Modal::EventDetail {
+                ref mut scroll_offset,
+                ..
+            } => {
+                *scroll_offset = scroll_offset.saturating_sub(1);
+                return;
+            }
             Modal::ModelPicker {
                 ref mut selected,
                 ref mut custom_active,
@@ -785,6 +779,14 @@ impl App {
 
     fn move_down(&mut self) {
         match self.state.modal {
+            Modal::EventDetail {
+                ref mut scroll_offset,
+                line_count,
+                ..
+            } => {
+                *scroll_offset = scroll_offset.saturating_add(1).min(max_scroll(line_count));
+                return;
+            }
             Modal::ModelPicker {
                 ref mut selected,
                 ref mut custom_active,
--- a/conductor-tui/src/input.rs
+++ b/conductor-tui/src/input.rs
@@ -78,8 +78,8 @@ pub fn map_key(key: KeyEvent, state: &AppState) -> Action {
         Modal::EventDetail { .. } => {
             return match key.code {
                 KeyCode::Esc | KeyCode::Char('q') => Action::DismissModal,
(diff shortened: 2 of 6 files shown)


devinrosen:
# Conductor PR Review

**0/4** reviewers approved — **blocking issues found**

Here is the architectural review:

---

## Architecture Review: `feat/141-feat-improve-agent-output`

### Files reviewed
`app.rs`, `state.rs`, `input.rs`, `ui/modal.rs`, `ui/mod.rs`, `ui/worktree_detail.rs`, `ui/help.rs`, `action.rs`

---

**Issue 1: Integer overflow on `horizontal_offset += 4`**
- **Severity**: warning
- **Location**: `conductor-tui/src/app.rs:172`
- **Details**: `*horizontal_offset += 4` on a `u16` will panic in debug builds after 16,383 right-scroll steps, and silently wrap in release. The symmetric left case correctly uses `saturating_sub`. Fix: `*horizontal_offset = horizontal_offset.saturating_add(4)`.

---

**Issue 2: Integer overflow on `*scroll_offset + 1`**
- **Severity**: warning
- **Location**: `conductor-tui/src/app.rs:324`
- **Details**: `(*scroll_offset + 1).min(max)` — if `scroll_offset` is `u16::MAX`, the addition overflows before `.min()` clamps it. Practically unlikely, but structurally inconsistent with how `saturating_sub` is used for the up direction. Fix: `scroll_offset.saturating_add(1).min(max)`.

---

**Issue 3: Duplicated source of truth — `line_count` stored alongside `body`**
- **Severity**: warning
- **Location**: `conductor-tui/src/state.rs:115`, `conductor-tui/src/ui/modal.rs:1072`
- **Details**: `Modal::EventDetail` stores both `body: String` and `line_count: usize`, where `line_count` is always `body.lines().count()`. This creates two sources of truth that must be kept in sync. Currently safe because `body` is never mutated after construction, but `render_event_detail` receives `line_count` as a separate parameter while ignoring `body.lines().count()`, making the API misleading. The render function should derive `line_count` from `body` locally — removing it from both the modal variant and the render function signature, reducing the parameter count from 7 to 6 and eliminating the redundancy.

---

**Issue 4: `AgentActivityDown`/`AgentActivityUp` repurposed for modal-internal scroll**
- **Severity**: suggestion...
*(truncated)*

---

## dry-abstraction — changes requested *(advisory)*

## DRY/Abstraction Review

### Issue 1: Duplicate title-truncation logic with inconsistent thresholds
**Severity**: warning
**Location**: `conductor-tui/src/ui/modal.rs` (render_event_detail, added lines ~1093–1100) vs `conductor-tui/src/app.rs:3498` (truncate_to_char_boundary)

`truncate_to_char_boundary` was added to `app.rs` and is used in `handle_expand_agent_event`. But `render_event_detail` re-implements title truncation inline:

```rust
let max_title_chars = (popup.width as usize).saturating_sub(7);   // truncates at width-7
let title_display = if title.chars().count() > (popup.width as usize).saturating_sub(4) {  // triggers at width-4
    let truncated: String = title.chars().take(max_title_chars).collect();
    format!(" {truncated}... ")
```

Two problems:
1. **Duplication**: `truncate_to_char_boundary` already exists but is `fn` (private), so `modal.rs` can't reuse it — causing re-implementation.
2. **Inconsistency**: The guard condition fires at `width - 4` chars but truncation takes `width - 7` chars, a mismatch of 3 characters. This creates a window where titles between `width-7` and `width-4` chars display with unnecessary `...`.

**Fix**: Make `truncate_to_char_boundary` `pub(crate)`, move it to a shared utility module (e.g., `conductor-tui/src/utils.rs`), and use it from `render_event_detail` to eliminate both issues.

---

### Issue 2: `line_count` field is a derived value cached in struct
**Severity**: suggestion
**Location**: `conductor-tui/src/state.rs:111`, `conductor-tui/src/app.rs:3255`

`Modal::EventDetail` stores both `body: String` and `line_count: usize`. The line count is trivially derivable from `body.lines().count()`. Storing both creates a risk of the two going out of sync if `body` is ever mutated in place.

**Fix**: Remove `line_count` from the struct and compute it in `render_event_detail` (`body.lines().count()`). The caller that sets `scroll_offset = line_count.saturating_sub(1)` (GoToBottom) can compute it from `body` too. ...
*(truncated)*

---

## security — changes requested *(required)*

## Security Review: feat/141-feat-improve-agent-output

### Issues Found

---

**Issue**: Unbounded `horizontal_offset` increment allows integer overflow  
**Severity**: suggestion  
**Location**: `conductor-tui/src/app.rs` — `Action::ScrollRight` handler  
**Details**: `*horizontal_offset += 4` with no upper bound. For a `u16`, this wraps on overflow in debug builds (panic) or wraps silently in release. Use `.saturating_add(4)` or cap at a reasonable max (e.g., max line width). The `scroll_offset` equivalent uses `.saturating_sub`, so this is inconsistent.

---

**Issue**: External process execution with user-controlled path (`log_path`)  
**Severity**: warning  
**Location**: `conductor-tui/src/app.rs:handle_copy_last_code_block` — `std::fs::File::open(log_path)`  
**Details**: `log_path` comes from `r.log_file` stored in the DB. If an attacker can write to the DB (e.g., via a compromised agent run record), they could point this at arbitrary files (e.g., `/etc/passwd`, SSH keys). This is an indirect read — no code execution risk here since the path is only opened for reading and the content is piped to `pbcopy`/`xclip`/`xsel`. However, the clipboard contents would then contain arbitrary file data exfiltrated silently. Mitigation: validate `log_path` is under the expected `~/.conductor/agent-logs/` directory before opening.

---

**Issue**: Code block content from log file piped directly to clipboard without sanitization  
**Severity**: suggestion  
**Location**: `conductor-tui/src/app.rs:handle_copy_last_code_block`  
**Details**: The extracted code block is written verbatim to the system clipboard. If an agent log contains a maliciously crafted "last code block" (e.g., from a prompt injection attack against the agent), this silently copies potentially dangerous commands to the user's clipboard for later execution. This is a low-severity clipboard poisoning risk inherent to the feature design — worth a user-visible warning or a confirmation step for large/...
*(truncated)*

---

## Performance Review

### Issues Found

---

**Issue**: `render_event_detail` rebuilds `Vec<Line>` from scratch on every render frame
**Severity**: warning
**Location**: `conductor-tui/src/ui/modal.rs` (render_event_detail function, ~line 1082)
**Details**: `body.lines().map(|l| Line::from(Span::raw(l.to_string()))).collect()` allocates a new `Vec<Line>` and `String` per line on every render tick. For large agent logs this could be hundreds or thousands of allocations per frame. The `body` string is already stored in `Modal::EventDetail` and doesn't change between frames. The pre-processed `Vec<Line>` should be cached in the modal state alongside `line_count`. Alternatively, `Span::raw` accepts `&str` (via `Into<Cow<str>>`), so `l.to_string()` can be replaced with `l` directly to avoid per-line `String` allocation, though the `Vec` rebuild per frame remains the core issue.

**Fix**: Either store `Vec<Line<'static>>` in `Modal::EventDetail` computed once at expand time, or use borrowed spans: `Line::from(l)` instead of `Line::from(Span::raw(l.to_string()))`.

---

**Issue**: Title character count computed twice with redundant `chars().count()` traversal
**Severity**: suggestion
**Location**: `conductor-tui/src/ui/modal.rs` (render_event_detail, title truncation block)
**Details**: `title.chars().count()` and `title.chars().take(max_title_chars).collect()` both walk the char iterator independently. For long titles this is two full traversals. Single pass: collect up to `max_title_chars` chars and check if iteration ended early to determine truncation. Also computed on every render frame.

---

**Issue**: `event_at_visual_index` does a linear scan of all events on every keypress
**Severity**: suggestion  
**Location**: `conductor-tui/src/state.rs:event_at_visual_index` (~line 348)
**Details**: O(n) walk through `agent_events` on each Up/Down or Expand action. For normal event counts (tens to low hundreds) this is fine. Only becomes a concern if event lists grow into t...
*(truncated)*

---

*Review cost: $0.5095 across 4 reviewers*

PR #195, opened 2026-03-05
fix: use agent_activity_len for ListState clamping in reload_agent_events

### LINKED ISSUE #194: TUI: Agent Activity cursor jumps when at last row after background refresh (multiple runs)
## Bug Description

In the Worktree Detail view, when the cursor is at the very last row of the Agent Activity list, it will jump upward by a number of rows after a few seconds (the background poll interval). This only occurs when the worktree has **multiple** agent runs in its history.

## Root Cause

In `conductor-tui/src/app.rs`, the `reload_agent_events()` function clamps the `ListState` selection using `agent_events.len()` (raw event count):

```rust
// line 588
let len = self.state.data.agent_events.len();
let cur = self.state.agent_list_state.borrow().selected();
if let Some(idx) = cur {
    if len == 0 {
        self.state.agent_list_state.borrow_mut().select(None);
    } else if idx >= len {
        self.state.agent_list_state.borrow_mut().select(Some(len - 1));
    }
}
```

However, the actual visual list rendered in `render_agent_activity()` uses `data.visual_rows()`, which **includes run-separator header rows** when there are multiple runs. The correct total is `data.agent_activity_len()`.

When there are N runs, N separator rows are inserted, so:
- `agent_activity_len()` = `agent_events.len()` + N separators
- Last visual index = `agent_events.len() + N - 1`

When the cursor is at the last visual row and the background poller fires `reload_agent_events()`:
1. `len = agent_events.len()` (unchanged or slightly larger)
2. `cur = agent_activity_len() - 1` (last visual row, which is > `agent_events.len() - 1`)
3. `cur >= len` → clamps to `len - 1`
4. Cursor jumps **up by N rows** (the number of run separator headers)

The "random seconds" timing corresponds to the background DB poll interval.

## Fix

Change the clamp in `reload_agent_events()` to use `data.agent_activity_len()` instead of `agent_events.len()`:

```rust
let len = self.state.data.agent_activity_len();
```

This must be called **after** `agent_events` and `agent_run_info` are both populated (since `agent_activity_len()` → `visual_rows()` uses both).

## Steps to Reproduce

1. Open a worktree that has ≥2 agent runs
2. Navigate to the last row in Agent Activity
3. Wait a few seconds for the background poller to fire
4. Observe cursor jumping up by the number of runs in history

devinrosen:
When multiple runs exist, the visual list includes run-separator header rows.
The ListState clamping was using agent_events.len() (raw event count) instead
of agent_activity_len() (visual row count including separators), causing the
cursor to jump up by N rows (number of run separators) on each background poll.

Fixes #194.

### DIFF
--- a/conductor-tui/src/app.rs
+++ b/conductor-tui/src/app.rs
@@ -585,7 +585,9 @@ impl App {
 
         // Clamp ListState selection to valid range after events reload.
         // ratatui also clamps during render, but we keep it tidy here.
-        let len = self.state.data.agent_events.len();
+        // Use agent_activity_len() (which includes run-separator rows) so the
+        // cursor isn't clamped below the last visual row when multiple runs exist.
+        let len = self.state.data.agent_activity_len();
         let cur = self.state.agent_list_state.borrow().selected();
         if let Some(idx) = cur {
             if len == 0 {


devinrosen:
# Conductor PR Review

**3/4** reviewers approved — all required checks passed

## architecture — approved *(required)*

**Architecture Review — feat/194-tui-agent-activity-cursor-jumps**

**Scope reviewed:** Coupling, cohesion, layer violations, design patterns.

**Analysis of `+` lines:**

The single meaningful change replaces direct field access (`agent_events.len()`) with a method call (`agent_activity_len()`). This is architecturally sound:

- **Encapsulation improved**: The visual-row count (which includes run-separator rows) is now computed in the data/state layer rather than being reconstructed ad-hoc in the TUI layer. This is the correct direction — the TUI should ask the model "how many rows are there?" rather than computing it from raw field access.
- **No layer violations**: TUI (`app.rs`) queries the state/data struct via a well-defined method. No DB calls, no business logic leaking into the view.
- **Cohesion**: The method name `agent_activity_len()` signals intent clearly and aligns with the existing naming pattern.

No issues found in `+` lines.

VERDICT: APPROVE

---

## DRY/Abstraction Review

**Reviewed:** `conductor-tui/src/app.rs` (lines 585–594 context)

### Added lines (+)

The only additions are:
1. Two comment lines clarifying why `agent_activity_len()` is used instead of `agent_events.len()`
2. The method call swap: `agent_events.len()` → `agent_activity_len()`

### Analysis

**No DRY or abstraction issues found.**

- The change moves from an inline `.len()` on the raw events slice to a dedicated `agent_activity_len()` method that accounts for visual separator rows. This is the correct direction — it centralises the "how many rows does the activity list have?" calculation, preventing future callers from needing to replicate the separator-counting logic themselves.
- No duplication is introduced. The new comment lines are appropriately scoped and non-redundant.
- No premature abstraction — `agent_activity_len()` represents a real, reusable concept (visual row count including separators) rather than a one-off wrapper.

---

VERDICT: APPROVE

---

## security — approved *(required)*

No security issues found.

The diff is a one-line logic fix replacing `agent_events.len()` with `agent_activity_len()` for ListState cursor clamping. The changed code:

- Reads a length value from an in-memory data structure
- Performs a comparison against a cursor index
- Contains no input validation, authentication, authorization, injection surface, or secrets

VERDICT: APPROVE

---

## Performance Review

**Diff scope:** The single changed `+` line swaps `agent_events.len()` (O(1), no allocation) for `agent_activity_len()`.

**`agent_activity_len()` implementation** (`state.rs:368-370`):
```rust
pub fn agent_activity_len(&self) -> usize {
    self.visual_rows().len()   // allocates a Vec<VisualRow> just to count
}
```

`visual_rows()` (`state.rs:341-364`) builds a `Vec` with capacity `events + runs`, iterates all events, and pushes rows — then `agent_activity_len` discards the Vec and returns only `.len()`.

**Is this a blocking issue for the diff?**

No. This call site is in a background-refresh path (`reload_agent_events`), not a per-frame render loop. The allocation is O(events), bounded by a realistic event count, and runs only when the DB is polled. The correctness fix (preventing cursor jumps) justifies the call; the allocation overhead is negligible in this context.

The underlying `agent_activity_len()` allocation-for-length pattern was already flagged in the prior review round as a non-blocking suggestion.

---

```
OFF-DIFF-FINDING
title: agent_activity_len() allocates a full Vec just to return .len()
file: conductor-tui/src/state.rs
line: 368
severity: suggestion
body: agent_activity_len() calls visual_rows().len(), which builds an owned Vec<VisualRow> 
(with capacity = events + runs) only to discard it after counting. 
This can be avoided by computing the count directly: 
  self.agent_events.len() + if self.agent_run_info.len() > 1 { self.agent_run_info.len() } else { 0 }
or by adding a separate iterator-based count path. Not blocking, but worth fixing if 
agent_activity_len() is called in a render loop in future.
END-OFF-DIFF-FINDING
```

---

`VERDICT: APPROVE`

---

## Off-Diff Findings (filed as issues)

The following issues were found in unchanged code and filed as separate GitHub issues:

- **agent_activity_len() allocates a full Vec just to return .len()** (`conductor-tui/src/state.rs`:368): [https://github.com/devinrosen/conductor-ai/issues/196](https://github.com/devinrosen/conductor-ai/issues/196) *(suggestion)*

---

*Review cost: $0.2067 across 4 reviewers*

PR #456, opened 2026-03-10
feat: add workflow resume command to resume failed/stalled runs

### LINKED ISSUE #445: feat: conductor workflow resume command
## Problem

When a workflow run fails mid-execution (e.g., hitting an API rate limit, transient network error, or resource exhaustion), there's no way to resume from the failed step. The only option is to re-run the entire workflow from scratch, which wastes time and compute — especially for workflows like `ticket-to-pr` that may have already completed expensive steps (plan, implement, push-and-pr) before failing.

## Context

The workflow engine is already designed for resumability:
- Each run stores a `definition_snapshot` so the engine never re-parses the `.wf` file
- `workflow_run_steps` tracks per-step status, iteration, and output
- The engine doc describes scanning for `running`/`waiting` runs and re-entering `execute_nodes()` from the last non-terminal step

The infrastructure is there — we just need the CLI/TUI/web entry points.

## Proposed solution

```
conductor workflow resume <run-id>
```

1. Load the workflow run and its `definition_snapshot` from DB
2. Validate the run is in `failed` or `running` status (not `completed`/`cancelled`)
3. Find the last step that succeeded and re-enter `execute_nodes()` from the next step
4. For steps inside a `while` loop, preserve the iteration counter
5. For `parallel` blocks where some agents completed, only re-run the failed/pending ones

### Edge cases
- If the failed step was inside a `while` loop, resume from that iteration (don't reset the counter)
- If the failed step was inside a `parallel` block, re-run only the agents that didn't complete
- A `--from-step <step-id>` flag could allow manually choosing the resume point
- A `--restart` flag could re-run from the beginning while reusing the same run record

### TUI/Web
- TUI: action on a failed workflow run to resume it
- Web: button on the workflow run detail page

## Motivation

Hit this in practice: a `ticket-to-pr` workflow run failed during the review swarm because of a rate limit, after plan + implement + push-and-pr had already completed successfully.

devinrosen:
## Summary
- Add `workflow resume` CLI subcommand with `--from-step` and `--restart` flags to resume failed/stalled workflow runs
- Add `POST /api/workflow-runs/:id/resume` web API endpoint for resuming runs
- Add TUI resume action (`r` key) for failed/interrupted workflow runs
- Core engine uses a skip-set in `ExecutionState` with composite `step_name:iteration` keys, checked by each node executor to skip already-completed steps

Closes #445

## Test plan
- [x] `cargo fmt --all --check` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes
- [ ] Manual: run a workflow, kill it mid-execution, then `conductor workflow resume <run-id>` to verify it picks up from last completed step
- [ ] Manual: test `--from-step` flag to resume from a specific step
- [ ] Manual: test `--restart` flag to restart the entire workflow

### DIFF
--- a/conductor-core/src/workflow.rs
+++ b/conductor-core/src/workflow.rs
@@ -1445,6 +1445,11 @@ pub fn validate_resume_preconditions(
                 .to_string(),
         ));
     }
+    if matches!(status, WorkflowRunStatus::Running) {
+        return Err(ConductorError::Workflow(
+            "Cannot resume a workflow run that is already running.".to_string(),
+        ));
+    }
     if matches!(status, WorkflowRunStatus::Cancelled) {
         return Err(ConductorError::Workflow(
             "Cannot resume a cancelled workflow run.".to_string(),
@@ -6453,6 +6458,16 @@ And here is my actual output:
         assert!(
             err.to_string().contains("Cannot resume a cancelled"),
             "Expected cancelled-run error, got: {err}"
+        );
+    }
+
+    #[test]
+    fn test_resume_rejects_running_run() {
+        let err =
+            validate_resume_preconditions(&WorkflowRunStatus::Running, false, None).unwrap_err();
+        assert!(
+            err.to_string().contains("already running"),
+            "Expected running-run error, got: {err}"
         );
     }
 


devinrosen:
# Conductor PR Review

**2/7** reviewers approved — **blocking issues found**

- :white_check_mark: **db-migrations** — approved *(advisory)*
- :white_check_mark: **security** — approved *(required)*

<details open>
<summary><strong>architecture</strong> — changes requested *(required)*</summary>

**Issue**: `ExecutionState` carries resume-specific fields that are always empty for fresh runs
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` — `ExecutionState` struct (around line 1028)
**Details**: `skip_completed: HashSet<String>` and `resume_step_map: HashMap<String, WorkflowRunStep>` are resume-only concerns embedded in the general execution state struct. For every fresh run these fields are initialized to empty collections that are never touched. This is a cohesion violation — `ExecutionState` now serves two responsibilities: tracking live execution state and carrying pre-loaded resume data. The direct symptom is the `std::mem::take` dance in `restore_step`:

```rust
fn restore_step(state: &mut ExecutionState<'_>, key: &str, iteration: u32) {
    let map = std::mem::take(&mut state.resume_step_map);
    restore_completed_step_from_db(state, &map, key, iteration);
    state.resume_step_map = map;
}
```

This borrow-splitting hack exists solely because the map is co-located with the mutable state it must update. The cleaner design is a separate `ResumeContext` (or `Option<ResumeContext>`) that is passed alongside `ExecutionState` through the execution functions that need it, keeping the core execution state free of resume infrastructure.

---

**Issue**: `run_workflow_engine` takes `workflow_run_id` and `parent_run_id` as explicit parameters that duplicate fields already present in `state`
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` — `run_workflow_engine` function signature (added around line 1163)
**Details**: `ExecutionState` already stores `state.workflow_run_id` and `state.parent_run_id`. The extracted `run_workflow_engine` function then takes these as separate `&str` parameters:

```rust
fn run_workflow_engine(
    state: &mut ExecutionState<'_>,
    workflow: &WorkflowDef,
    workflow_run_id: &str,   // already in state.workflow_run_id
    parent_run_id: &str,     // already in state.parent_run_id
) -...
*(truncated)*

</details>

---

<details open>
<summary><strong>dry-abstraction</strong> — changes requested *(advisory)*</summary>

**Issue**: `get_completed_step_keys` is defined but bypassed — `resume_workflow` inlines the same logic
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` — `get_completed_step_keys` method and `resume_workflow` function
**Details**: `get_completed_step_keys` was added as a public helper to compute the skip set, but `resume_workflow` never calls it. Instead, it recomputes the exact same logic inline using `all_steps`:

```rust
// In resume_workflow (inline):
let mut keys: HashSet<String> = all_steps
    .iter()
    .filter(|s| s.status == WorkflowStepStatus::Completed)
    .map(|s| format!("{}:{}", s.step_name, s.iteration))
    .collect();
```

```rust
// In get_completed_step_keys (unused for resume):
for step in &steps {
    if step.status == WorkflowStepStatus::Completed {
        keys.insert(format!("{}:{}", step.step_name, step.iteration));
    }
}
```

`resume_workflow` skips `get_completed_step_keys` to avoid a second DB fetch (since `all_steps` is already loaded), but the transformation logic is duplicated. Extract a free function `fn step_keys_from_steps(steps: &[WorkflowRunStep]) -> HashSet<String>` that both the helper method and `resume_workflow` can call.

---

**Issue**: Step skip-key format string `"{}:{}"` repeated ~10 times across the diff
**Severity**: warning
**Location**: Multiple locations in `conductor-core/src/workflow.rs`
**Details**: The pattern `format!("{}:{}", step_name, iteration)` for constructing skip-set lookup keys appears in at least 10 `+` lines:

- `get_completed_step_keys`: `format!("{}:{}", s.step_name, s.iteration)`
- `resume_workflow` inline key build: `format!("{}:{}", s.step_name, s.iteration)`
- `resume_workflow` `to_remove` build: `format!("{}:{}", s.step_name, s.iteration)`
- `resume_workflow` `resume_step_map` build: `format!("{}:{}", s.step_name, s.iteration)`
- `execute_call_with_schema`: `format!("{}:{}", step_key_check, iteration)`
- `execute_parallel`: `format!("{}:{}", agent_step_key, iterati...
*(truncated)*

</details>

---

<details open>
<summary><strong>error-handling</strong> — changes requested *(advisory)*</summary>

*Focus: Unwraps, swallowed errors, vague messages, missing context*

**Issue**: Silent no-op when a step is in `skip_completed` but not found in `resume_step_map`
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` — `restore_completed_step_from_db` (the `if let Some(step) = completed_step` block)
**Details**: When `restore_step` is called (triggered because `skip_completed.contains(&skip_key)` is true), and the lookup against `resume_step_map` comes up empty, the function silently returns without logging anything. The call site then returns `Ok(())` as if the step was successfully skipped and its state restored. In practice, `step_results` and `contexts` are not populated for that step, which means downstream steps using `{{prior_context}}` or checking markers will silently operate on missing data rather than failing with a clear message. The missing log makes this nearly impossible to diagnose. A `tracing::warn!` like `"resume: step '{step_key}:{iteration}' in skip set but not found in resume_step_map — downstream variable substitution may be incorrect"` should be emitted before returning.

---

**Issue**: Error message for missing `definition_snapshot` omits the run ID
**Severity**: suggestion
**Location**: `conductor-core/src/workflow.rs` — `resume_workflow` snapshot extraction
**Details**: `"Workflow run has no definition snapshot — cannot resume."` doesn't include `wf_run.id`. When this error surfaces in logs or CLI output, the operator has to correlate back from context. Prefer `format!("Workflow run '{}' has no definition snapshot — cannot resume.", wf_run.id)`.

---

**Issue**: `--from-step` not-found error omits the run ID
**Severity**: suggestion
**Location**: `conductor-core/src/workflow.rs` — `resume_workflow` `from_step` resolution
**Details**: `"Step '{}' not found in workflow run"` gives the step name but not the run ID. Including the run ID (`wf_run.id`) would let operators quickly look up the run record.

VERDICT: REQUEST_CHANGES

</details>

---

<details open>
<summary><strong>performance</strong> — changes requested *(advisory)*</summary>

**Issue**: Double `get_workflow_steps` DB query in `resume_workflow`
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` (in `resume_workflow`, the `resume_step_map` construction)
**Details**: `all_steps` is already loaded earlier in the function (with the comment "Load all steps once (avoids N+1 queries later)"), but `resume_step_map` is built by calling `wf_mgr.get_workflow_steps(&wf_run.id)?` a second time. The re-fetch justification is "after resets", but `reset_failed_steps` only clears non-completed steps and `reset_steps_from_position` only clears steps whose keys were already removed from `skip_completed`. The `resume_step_map` filters exclusively for `Completed` steps, which are unchanged by those resets. `all_steps` can be reused directly, filtered by membership in `skip_completed`:
```rust
resume_step_map: all_steps
    .into_iter()
    .filter(|s| skip_completed.contains(&format!("{}:{}", s.step_name, s.iteration)))
    .map(|s| (format!("{}:{}", s.step_name, s.iteration), s))
    .collect(),
```

---

**Issue**: N+1 DB queries when accumulating costs from completed steps during resume
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs`, `restore_completed_step_from_db`, inside `if let Some(ref child_run_id) = step.child_run_id`
**Details**: `state.agent_mgr.get_run(child_run_id)` issues one DB query per completed step. For a workflow with 20 completed steps, this is 20 sequential round-trips. The `child_run_id`s are already present in the step records loaded into `resume_step_map`. They should be batch-loaded once before entering the execution loop — e.g., collect all non-null `child_run_id` values from `resume_step_map`, issue a single `WHERE id IN (...)` query, and store the results in a `HashMap<String, AgentRun>` that `restore_completed_step_from_db` reads from instead of querying per-step.

---

**Issue**: Unconditional `String` allocation for `skip_key` on every agent step execution
**Severity**: suggestion
**Lo...
*(truncated)*

</details>

---

<details open>
<summary><strong>test-coverage</strong> — changes requested *(advisory)*</summary>

*Focus: Missing tests for new behavior, bug fixes, and public APIs*

**Issue**: `execute_gate` skip path returns `Ok(())` without calling `restore_step()`, unlike all other node types
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` (gate skip block, diff line ~+2672–2679)
**Details**: When a gate is in `skip_completed` on resume, the code logs and returns immediately without restoring gate state into `state.step_results` or `state.contexts`. Every other node type (call, parallel, sub-workflow) calls `restore_step()` after the skip check to rebuild execution state. For gates this means: (a) the gate step is absent from `step_results`, breaking any downstream `{{step_results.gate_name}}` reference; (b) `state.last_gate_feedback` stays empty, so downstream steps using `{{gate_feedback}}` get an empty string instead of the stored feedback. No test covers gate skip on resume at all.

---

**Issue**: `resume_workflow()` — no test for missing `definition_snapshot` error path
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` (snapshot deserialization block, diff lines +~1330–1338)
**Details**: `resume_workflow` returns an error when `definition_snapshot` is `None`. Runs created before snapshotting was added (or with a serialization bug) would hit this path in production. There is no test that creates a `WorkflowRun` without a snapshot and verifies the correct `ConductorError::Workflow` is returned. The visible test suite covers `Completed`/`Cancelled` rejection but not this third validation branch.

---

**Issue**: `find_max_completed_while_iteration()` silently skips `Parallel` body nodes — no test for while+parallel combination
**Severity**: suggestion
**Location**: `conductor-core/src/workflow.rs` (find_max_completed_while_iteration, diff lines +~3023–3060)
**Details**: The function filters body nodes to only `Call` nodes; `Parallel` branches return `None` and are discarded. If a while body contains a parallel block, `body_keys` will be empty (or partial), causing the function to return 0 ...
*(truncated)*

</details>

---

*Review cost: $0.8325 across 7 reviewers*

devinrosen:
# Conductor PR Review

**4/7** reviewers approved — **blocking issues found**

- :white_check_mark: **db-migrations** — approved *(advisory)*
- :white_check_mark: **dry-abstraction** — approved *(advisory)*
- :white_check_mark: **error-handling** — approved *(advisory)*
- :white_check_mark: **security** — approved *(required)*

<details open>
<summary><strong>architecture</strong> — changes requested *(required)*</summary>

**Issue**: `find_max_completed_while_iteration` silently falls back to iteration 0 for while bodies containing only `Parallel`, `Gate`, or `CallWorkflow` nodes
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` (the `find_max_completed_while_iteration` function, approx. the block starting at `let body_keys: Vec<String> = node.body.iter()...`)
**Details**: The function only extracts step keys from `WorkflowNode::Call` variants, returning `None` for `Parallel`, `Gate`, and `CallWorkflow`. If a while body is composed entirely of those node types, `body_keys` will be empty and the function returns 0. The caller interprets 0 as "no iterations completed" and restarts the loop from the beginning, silently discarding all completed iterations and re-running potentially expensive work. The fix is to recurse into all trackable node variants (parallel agents, gates, sub-workflow calls) to find their step keys, or to query `workflow_run_steps` directly by `parallel_group_id`/position range for the while body's position span rather than reconstructing keys from the node tree.

---

**Issue**: `restore_step` uses a `.take()`/put-back pattern to split borrows on `ExecutionState`
**Severity**: suggestion
**Location**: `conductor-core/src/workflow.rs` (`restore_step` function)
**Details**: Temporarily moving `resume_ctx` out of `state` to satisfy the borrow checker is a code smell that signals a structural coupling problem: `restore_completed_step` needs both `&mut ExecutionState` and `&ResumeContext` simultaneously, but both live inside `ExecutionState`. The clean solution is to move the resume context out of `ExecutionState` entirely — pass `Option<&ResumeContext>` as a separate parameter to the functions that need it, rather than embedding it inside the mutable state. This would eliminate the borrow dance and make the data flow explicit at call sites.

---

**Issue**: `PLAN.md` committed as a permanent file in the repository root
**Severity**: suggestion
**Loca...
*(truncated)*

</details>

---

<details open>
<summary><strong>performance</strong> — changes requested *(advisory)*</summary>

**Issue**: N+1 DB queries when batch-loading child agent runs
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs` — `resume_workflow`, the `child_runs` HashMap construction
**Details**: The code comment says "Batch-load child agent runs to avoid N+1 queries" but the implementation issues one `agent_mgr.get_run(id)` call per completed step with a child run — N individual `SELECT` queries instead of one `SELECT … WHERE id IN (…)`. The pre-loading trades N+1 during `restore_completed_step` for N+1 during the initial load, so the total query count is unchanged.

```rust
// Current — N queries:
let child_runs: HashMap<String, crate::agent::AgentRun> = step_map
    .values()
    .filter_map(|s| s.child_run_id.as_deref())
    .filter_map(|id| match agent_mgr.get_run(id) { … })
    .collect();
```

The fix is to add a `get_runs_by_ids(&[&str]) -> Result<Vec<AgentRun>>` method to `AgentManager` that executes a single `SELECT … WHERE id IN (…)` query using a comma-joined placeholder list, then collect the results into the HashMap.

---

**Issue**: String allocations in `find_max_completed_while_iteration` inner loop
**Severity**: suggestion
**Location**: `conductor-core/src/workflow.rs` — `find_max_completed_while_iteration`
**Details**: `make_step_key(k, iter)` is called for every body key on every loop iteration, producing one heap-allocated `String` per check. For a while loop with M body nodes and I completed iterations that's M×I transient allocations. Pre-compute the keys per iteration outside the `all()` call, or cache the iteration-specific keys as the loop advances.

---

**Issue**: Dynamic SQL string allocation in reset helper methods
**Severity**: suggestion
**Location**: `conductor-core/src/workflow.rs` — `reset_failed_steps`, `reset_completed_steps`, `reset_steps_from_position`
**Details**: Each method calls `format!(…, Self::RESET_STEP_SET)` to produce a `String` from two compile-time-known string literals. Since all three...
*(truncated)*

</details>

---

<details open>
<summary><strong>test-coverage</strong> — changes requested *(advisory)*</summary>

*Focus: Missing tests for new behavior, bug fixes, and public APIs*

**Issue**: `resume_workflow()` has no test for the `--from-step` error path when the specified step name does not exist in the run.
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs:1415-1423`
**Details**: The `from_step` branch executes `all_steps.iter().find(|s| s.step_name == from_step)` and returns `Err` when `from_pos` is `None`. All four `resume_workflow` tests use `from_step: None`. The user-facing error message path ("Step '{}' not found in workflow run '{}'") is completely untested. Since this is a public function with a distinct error variant, a simple DB-level unit test (no mock agents needed) should cover it.

---

**Issue**: `resume_workflow()` has no test for the `restart: true` path.
**Severity**: warning
**Location**: `conductor-core/src/workflow.rs:1382-1391`
**Details**: When `restart: true`, the function calls both `reset_failed_steps()` and `reset_completed_steps()` and produces an empty `skip_completed` set (so `resume_ctx` is `None`). None of the four `resume_workflow` tests exercise `restart: true`. The DB-level side-effects (all steps become pending, resume_ctx is None) are testable without running agents — `get_workflow_steps()` can assert the post-state just as `test_reset_completed_steps` does.

---

**Issue**: `resume_workflow_endpoint` (web API) allows `restart: true` on a completed run but `resume_workflow()` core unconditionally rejects completed runs regardless of the restart flag.
**Severity**: warning
**Location**: `conductor-web/src/routes/workflows.rs:281`, `conductor-core/src/workflow.rs:1364-1370`
**Details**: The web endpoint's pre-validation gate is `if matches!(run.status, Completed) && !req.restart.unwrap_or(false)` — meaning a request with `restart: true` on a completed run passes pre-validation and gets a 202 ACCEPTED response. However, `resume_workflow()` (called from the spawned blocking task) unconditionally returns `Err` for completed runs before inspecting the restart flag. The background task si...
*(truncated)*

</details>

---

*Review cost: $1.0601 across 7 reviewers*

Issue #462, opened 2026-03-10
execute_while body loop may be missing fail_fast check


devinrosen:
**Severity**: warning
**File**: `conductor-core/src/workflow.rs`:0
**Found by**: dry-abstraction reviewer
**Source PR branch**: `feat/459-feat-add-do-while-loop-construct-to`

## Details

execute_do_while's body loop checks `!state.all_succeeded && state.exec_config.fail_fast` and returns early. The unchanged body loop in execute_while does not appear to have this guard (it was not modified in this PR), creating an inconsistency: a prior step failure will short-circuit do-while loops but not while loops. Both should use the same body-execution helper.

---
*Filed automatically by Conductor PR review swarm. This issue was found in unchanged code during a PR review and does not block the originating PR.*

PR #470, opened 2026-03-10
test: add test_while_fail_fast_exits_early to cover execute_while body loop (#462)

devinrosen:
Mirrors test_do_while_fail_fast_exits_early — verifies that the
fail_fast guard in the execute_while body loop exits early with Ok(())
when a prior failure has occurred.

### DIFF
--- a/conductor-core/src/workflow.rs
+++ b/conductor-core/src/workflow.rs
@@ -4445,6 +4445,42 @@ And here is my actual output:
     }
 
     #[test]
+    fn test_while_fail_fast_exits_early() {
+        let conn = setup_db();
+        let config = Config::default();
+        let mut state = make_loop_test_state(&conn, &config);
+        state.exec_config.fail_fast = true;
+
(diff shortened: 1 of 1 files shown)


devinrosen:
## Conductor Review Swarm — All Clear

| Reviewer | Verdict |
|----------|---------|
| architecture | :white_check_mark: approve |
| security | :white_check_mark: approve |
| performance | :white_check_mark: approve |
| error-handling | :white_check_mark: approve |
| db-migrations | :white_check_mark: approve |
| dry-abstraction | :white_check_mark: approve |
| test-coverage | :white_check_mark: approve |

All reviewers passed with no blocking findings.

### Suggestions (non-blocking)
- **test-coverage**: The body loop checks `fail_fast` after executing each node (unlike `execute_nodes` which checks before), so one body node always runs before the guard fires. The test uses a no-op `IfNode` that would pass even if the check order changed — a test with a real side-effecting body node would give stronger coverage of the semantic.

Issue #457, opened 2026-03-10
bug: closing TUI kills workflow executor thread, leaving steps stuck in running


devinrosen:
## Problem

The workflow executor runs as a `std::thread::spawn` inside the TUI process (`conductor-tui/src/app.rs:4302`). When the TUI is closed (intentionally or accidentally), the process exits and the executor thread is killed mid-execution. This leaves workflow steps stuck in `running` status even though the underlying agent run completed successfully.

**Observed:** `ticket-to-pr` workflow on `feat-445-feat-conductor-workflow-resume` — the `address-reviews` step is stuck at `running` (position 11, iteration 0) while its child agent run (`01KKAQ8HC39SJPTCX156N85Q1E`) completed with valid `CONDUCTOR_OUTPUT` and status `completed`.

## Root cause

`spawn_workflow_in_background()` uses `std::thread::spawn` — the thread has no lifecycle management. When the TUI process exits:
1. The thread is killed wherever it happens to be (polling, updating DB, spawning agents)
2. No cleanup runs — step status is not updated, no `failed`/`cancelled` marker is written
3. The workflow run stays `running` forever in the DB

## Proposed solutions

**Option A — Graceful shutdown:** On TUI exit, signal the workflow thread to stop (e.g., via `AtomicBool` or channel) and wait for it to reach a safe checkpoint before exiting. Steps in progress would be marked `failed` with "TUI closed".

**Option B — Out-of-process executor:** Run workflows in a separate process (tmux session or background daemon) that survives TUI restarts. The TUI would just monitor progress via DB polling. This aligns with the v2 daemon extraction plan.

**Option C — Startup recovery:** On TUI startup, scan for workflow steps stuck in `running` where the child agent run is terminal (`completed`/`failed`). Sync the step status from the agent run. This is complementary to A or B and also helps with #445 (workflow resume).

Option A is the minimum viable fix. Option C is valuable regardless of which primary approach is chosen.

## Related

- #445 — `conductor workflow resume` command (would allow recovering from this state)
- #447 — workflow `eprintln!` bleed-through (same root cause: workflow runs in TUI thread)

PR #479, opened 2026-03-10
fix(tui): graceful workflow shutdown and startup recovery on TUI exit (#457)

devinrosen:
Two complementary fixes for steps stuck in 'running' when TUI closes:

A) Graceful shutdown: `WorkflowExecConfig` gains an optional
   `Arc<AtomicBool>` shutdown flag that is checked on every poll
   iteration in `poll_child_completion`. When the TUI exits,
   `App::run()` sets the flag, then joins all workflow threads with a
   10-second bounded timeout. In-flight steps are marked `failed` with
   "workflow cancelled: TUI was closed" and the executor returns an
   error that aborts the workflow immediately.

B) Startup recovery: `WorkflowManager::recover_stuck_steps()` scans
   for `workflow_run_steps` with `status = 'running'` and a non-null
   `child_run_id`. For each, it loads the child `agent_run`; if it is
   terminal (completed/failed/cancelled), the step status is synced.
   Called from the background DB-poll throttle block alongside the
   existing orphan reaper.

### DIFF
--- a/conductor-tui/src/background.rs
+++ b/conductor-tui/src/background.rs
@@ -55,10 +55,10 @@ pub fn poll_data() -> Option<Action> {
             let _ = agent_mgr.reap_orphaned_runs();
             let _ = wt_mgr.reap_stale_worktrees();
             let wf_mgr = conductor_core::workflow::WorkflowManager::new(&conn);
-            if let Ok(n) = wf_mgr.recover_stuck_steps() {
-                if n > 0 {
-                    tracing::debug!("Recovered {n} stuck workflow step(s)");
-                }
+            match wf_mgr.recover_stuck_steps() {
+                Ok(n) if n > 0 => tracing::debug!("Recovered {n} stuck workflow step(s)"),
+                Ok(_) => {}
+                Err(e) => tracing::warn!("recover_stuck_steps failed: {e}"),
             }
         }
     }


devinrosen:

| Reviewer | Verdict |
|----------|---------|
| db-migrations | :white_check_mark: approve |
| security | :white_check_mark: approve |
| error-handling | :x: changes requested |
| test-coverage | :x: changes requested |
| dry-abstraction | :x: changes requested |
| architecture | :x: changes requested |
| performance | :x: changes requested |

### Blocking findings

<details>
<summary><b>error-handling</b> — 1 issue</summary>

- **warning** `conductor-core/src/workflow/manager.rs` — `recover_stuck_steps` silently discards DB update errors and over-counts recovered steps; errors from `update_step_status_full` should be propagated or at minimum logged
</details>

<details>
<summary><b>test-coverage</b> — 2 issues</summary>

- **warning** `conductor-core/src/workflow/manager.rs` — `recover_stuck_steps()` is a new public method with non-trivial DB query and status-mapping logic but has zero test coverage
- **warning** `conductor-tui/src/` — the new shutdown flag path in `poll_child_completion()` is never exercised; all existing tests pass `None` for the shutdown flag
</details>

<details>
<summary><b>dry-abstraction</b> — 1 issue</summary>

- **warning** `conductor-core/src/workflow/` — fragile string-prefix protocol (`"shutdown:"`) is used to communicate shutdown vs. normal poll failure between `poll_child_completion` and its caller; if either side drifts, shutdown errors will be silently misclassified as normal failures and retried — extract prefix and message strings to shared constants
</details>

<details>
<summary><b>architecture</b> — 2 issues</summary>

- **warning** `conductor-core/src/workflow/` — TUI-specific error messages (e.g. referencing TUI behaviour) leak into the core library crate, violating the library-first architecture; error messages in core should be UI-agnostic
- **warning** `conductor-core/src/workflow/` — string-prefix error discrimination is fragile; replace with a typed error enum variant to distinguish shutdown from other poll errors
</details>

<details>
<summary><b>performance</b> — 1 issue</summary>

- **warning** `conductor-core/src/workflow/manager.rs` — N+1 query pattern in `recover_stuck_steps()`: performs 2N redundant DB queries (including unnecessary plan-step fetches) per recovery cycle; batch the queries instead
</details>

### Suggestions (non-blocking)

- **error-handling**: Log `get_run` errors separately from the expected `Ok(None)` case to aid debugging
- **dry-abstraction**: Add a narrower helper for the `update_step_status_full` call pattern repeated in `recover_stuck_steps`
- **architecture**: Add structured logging in the recovery path; consider a helper to join threads with a timeout rather than sequential blocking joins
- **performance**: Use `Relaxed` instead of `SeqCst` for the shutdown flag load in the tight poll loop; optimize the thread-join loop to reap already-finished threads before blocking on slow ones

Issue #1211, opened 2026-03-18
bug/feat: Workflow Runs tree shows all loop iterations — should show only the latest


devinrosen:
## Problem

The Workflow Runs panel currently shows every loop iteration of a looping workflow (e.g., `iterate-pr`) as fully expanded step trees. After 2 loops, you see two complete copies of lint-fix → rebase-worktree → review-pr. This is noisy and makes it hard to focus on what's currently happening.

## Desired Behavior

**Always show only the latest iteration's child runs**, regardless of whether the run is in progress or complete:

- If a workflow completed after 3 loops → show only loop 3's steps/child runs
- If a workflow is on loop 2 (loop 1 done, loop 2 running) → show only loop 2's child runs
- If loop 2 fails and loop 1 completed → still show only loop 2 (latest), since that's what needs attention

No expand/collapse for prior iterations — users can inspect the workflow details view for full history.

## Iteration Badge

When the run is past its first iteration, show an **"Nth iteration"** badge on the parent run row so users know how many loops have occurred. Examples:
- `○ iterate-pr  —  2nd iteration`
- `✓ iterate-pr  4m12s  3rd iteration`

No badge needed for the first iteration (iteration 0). (`push_iteration_badge()` in `workflows.rs` already does this partially.)

---

## Root Cause (confirmed via code analysis)

The bug is in `visible_workflow_run_rows()` in `conductor-tui/src/state.rs`. The `children_map` (lines 1433–1443) is built purely from `parent_workflow_run_id` relationships with **no iteration awareness**:

```rust
for run in runs {
    if let Some(ref parent_id) = run.parent_workflow_run_id {
        if known_ids.contains(parent_id.as_str()) {
            children_map.entry(parent_id.as_str()).or_default().push(run);
        }
    }
}
```

For `iterate-pr` running 2 loops, both loops spawn child runs (`lint-fix`, `rebase-worktree`, `review-pr`) with the same `parent_workflow_run_id`. All 6 child runs land in `children_map` and all 6 get rendered by `push_children`.

The existing `collapse_loop_iterations()` function (in `app/helpers.rs`) only operates on `WorkflowRunStep` rows within a single run — it has **no effect on child `WorkflowRun` entries**. So it's working correctly, just at the wrong layer.

**The complication:** `WorkflowRun` has no `iteration` field. The only place the iteration↔child_run mapping exists is in the parent run's `workflow_run_steps` table via `(step_name, iteration, child_run_id)` tuples.

[...]

PR #1220, opened 2026-03-18
fix: fetch steps for all runs and interleave direct-call steps in tree

### LINKED ISSUE #1217: bug: direct `call` steps on non-leaf workflow parents are invisible in run tree
## Summary

In `iterate-pr` (and any workflow that mixes `call workflow` sub-workflows with direct `call` agent steps at the same level), direct agent call steps like `address-reviews` never appear in the TUI workflow run tree.

## Root cause

Two compounding issues:

**1. `background.rs` only batch-fetches steps for leaf runs**

\`\`\`rust
// background.rs ~line 607
let runs_with_children = runs.iter()
    .filter_map(|r| r.parent_workflow_run_id.as_deref())
    .collect();
let leaf_run_ids = runs.iter()
    .filter(|r| !runs_with_children.contains(r.id.as_str()))  // ← non-leaves excluded
    ...
let all_run_steps = wf_mgr.get_steps_for_runs(&leaf_run_ids);
\`\`\`

`iterate-pr` has child runs (`rebase-worktree`, `lint-fix`, `review-pr`), so it is classified as non-leaf. Its own steps are never loaded into `all_run_steps`, which drives the tree view.

**2. The tree renderer never calls `push_steps_for_run` on a non-leaf parent**

\`\`\`rust
// state.rs ~line 1534
if child_count == 0 {
    push_steps_for_run(run.id, ...);  // only leaf runs get steps shown
} else {
    push_children(run.id, ...);       // only sub-workflow child runs shown
}
\`\`\`

`address-reviews` is a direct `call` (agent call), not `call workflow`, so it creates a step in `iterate-pr`'s `workflow_run_steps` but no child run record. It never appears in `push_children` (which walks `workflow_runs` records only), and `push_steps_for_run` is never called for the parent.

## Affected workflow

```
// iterate-pr.wf (simplified)
do {
  call workflow rebase-worktree    ← creates child run ✓ visible
  call workflow review-pr          ← creates child run ✓ visible
  if review-aggregator.has_blocking_findings {
    call address-reviews           ← only a step, no child run ✗ invisible
    script fmt-fix { ... }         ← same ✗ invisible
    script push { ... }            ← same ✗ invisible
    call workflow lint-fix         ← creates child run ✓ visible
  }
} while review-aggregator.has_blocking_findings
```

## Why this is specific to `iterate-pr`

Most workflows are either all sub-workflow calls (non-leaf, all children visible) or flat agent calls (leaf, all steps visible). `iterate-pr` mixes both at the same level, which exposes the gap.

## Proposed fix

Either:
- **Option A**: In `background.rs`, also batch-fetch steps for non-leaf runs that have at least one direct `call` step (i.e., steps without a `child_run_id`). Render those steps inline in the tree alongside sub-workflow children.
- **Option B**: In the tree, when a non-leaf parent is selected/expanded, also emit step rows for its non-sub-workflow steps.

devinrosen:
## Problem

Two related bugs caused by non-leaf workflow runs (runs with children, like `iterate-pr`) never having their steps fetched:

**Bug 1 — Loop iteration filter silently no-ops (showing all iterations)**

The #1211 fix built `latest_child_ids` from `workflow_run_steps.get(parent_id)`, but non-leaf parents were excluded from the batch step fetch in `background.rs`. So `latest_child_ids` was always empty → guard condition `!latest_child_ids.is_empty()` was always false → all loop iterations were shown instead of just the latest.

**Bug 2 — Direct `call` agent steps invisible in tree (closes #1217)**

Steps without a `child_run_id` (e.g. `address-reviews`, `fmt-fix`, `push` in `iterate-pr`) belong to the parent run's step records. Since those records were never fetched for non-leaf runs, these steps were completely invisible in the tree.

### `background.rs`
Remove the leaf-only filter — fetch steps for **all** runs via the existing batch `get_steps_for_runs` query (single `IN` clause, max 50 runs, no performance concern).

### `state.rs` — `push_children`
Rewrite to merge child runs and direct-call steps into a single position-sorted list:
1. `latest_child_ids` — now works correctly (parent step data is available)
2. `child_run_position` map — `child_run_id → position` for sort ordering
3. `direct_steps` — parent's own steps with `child_run_id.is_none()` at latest iteration
4. Merged + sorted by position → emitted in execution order

Direct steps are only shown when the parent run is in `expanded_step_run_ids`.

## Result

`iterate-pr` tree now shows (latest iteration only, in order):
- `rebase-worktree` [child run]
- `review-pr` [child run]
- `address-reviews` [direct step] ← was invisible
- `fmt-fix` [direct step] ← was invisible
- `push` [direct step] ← was invisible
- `lint-fix` [child run]

## Tests
3 new tests in `state.rs`:
- `non_leaf_parent_direct_steps_interleaved_with_children`
- `non_leaf_parent_direct_steps_hidden_when_not_expanded`
- `non_leaf_parent_loop_filter_and_direct_steps`

All 188 tests pass.

### DIFF
--- a/conductor-tui/src/background.rs
+++ b/conductor-tui/src/background.rs
@@ -608,13 +608,37 @@ fn poll_workflow_data(
     //   2. Direct agent-call steps (steps without a child_run_id) on non-leaf parents are
     //      visible in the tree alongside their sub-workflow children.
     let all_run_ids: Vec<&str> = runs.iter().map(|r| r.id.as_str()).collect();
-    let all_run_steps = match wf_mgr.get_steps_for_runs(&all_run_ids) {
+    let known_run_id_set: std::collections::HashSet<&str> = all_run_ids.iter().copied().collect();
+
+    // In global/repo mode the run list is capped at 50. A parent run (e.g. iterate-pr)
+    // that started before its many children may fall outside that cap while its children
+    // are still visible. Without the parent's step records, push_children cannot build
+    // latest_child_ids and shows all loop iterations. Collect any parent IDs referenced
+    // by visible runs that are themselves outside the cap, and fetch their steps too.
+    let ancestor_ids: Vec<&str> = runs
+        .iter()
+        .filter_map(|r| r.parent_workflow_run_id.as_deref())
+        .filter(|pid| !known_run_id_set.contains(pid))
+        .collect::<std::collections::HashSet<_>>()
+        .into_iter()
+        .collect();
+
+    let mut all_run_steps = match wf_mgr.get_steps_for_runs(&all_run_ids) {
         Ok(steps) => steps,
         Err(e) => {
             tracing::warn!("get_steps_for_runs failed for runs {:?}: {e}", all_run_ids);
             Default::default()
         }
     };
+    if !ancestor_ids.is_empty() {
+        match wf_mgr.get_steps_for_runs(&ancestor_ids) {
+            Ok(ancestor_steps) => all_run_steps.extend(ancestor_steps),
+            Err(e) => tracing::warn!(
+                "get_steps_for_runs failed for ancestor runs {:?}: {e}",
+                ancestor_ids
+            ),
+        }
+    }
 
     // Load agent events for the selected step's child run
     let agent_mgr = AgentManager::new(&conn);


devinrosen:
Superseded by #1227 which stores iteration directly on the WorkflowRun record instead of cross-referencing parent step records.

PR #1776, opened 2026-04-04
feat: per-step estimation, live narrowing, confidence intervals, TUI display

alabele:
## Summary

Extends the v1 hybrid time estimation system with four capabilities:

- **Per-step estimation** — new `get_completed_step_durations` query fetches historical step durations grouped by `(workflow_name, step_name)`, enabling individual step estimates via `estimate_all_steps()`
- **Live narrowing** — `live_remaining_estimate()` sums per-step estimates for incomplete steps. Running steps subtract elapsed time; completed steps contribute 0. Replaces the static `total - elapsed` approach.
- **Confidence intervals** — `Estimate` struct carries p25/p75 bounds and a `Confidence` level (High/Medium/Low) derived from sample size and coefficient of variation. Ranges (`~8-14m left`) shown when confidence is medium/high; point estimates for low confidence.
- **TUI display** — running workflow rows show `~Xm left` or `~X-Ym left` instead of `…`, computed in the background poller and stored in `DataCache`

## Test plan

- [x] `cargo build` — all crates compile
- [x] `cargo test -p conductor-core --features test-helpers -- estimation` — 29 unit tests pass
- [x] `cargo clippy -- -D warnings` — clean
- [x] Frontend builds (`bun run build`)
- [ ] Manual: `GET /api/workflows/runs?status=running` returns `estimate_confidence`, `estimated_remaining_low_ms`, `estimated_remaining_high_ms`, `step_estimates`
- [ ] Manual: TUI shows `~X-Ym left` for runs with enough historical data
- [ ] Manual: StepDetailPanel shows per-step estimate for non-completed steps


devinrosen (commented):
1. Confidence aggregation bug in live_remaining_estimate (estimation.rs:~195)                                                                                                                                                                                                
                                                                                                                                                                                                                                                                               
  The worst-confidence aggregation iterates over step_estimates.values() — all steps in the map — rather than only the steps that actually contributed to the remaining time. If a workflow has 10 high-confidence completed steps and 1 low-confidence pending step that has  
  no estimate, the entire display shows Low confidence unnecessarily. The aggregation should only consider estimates for steps that passed the has_any = true gate.                                                                                                            
                  
  2. Stale watchdog is dead code in this PR (recovery.rs)                                                                                                                                                                                                                      
                  
  detect_stale_workflow_runs and reap_stale_workflow_runs are added and exported, but nothing in the background poller, CLI, or web server calls them. The fire_stale_workflow_notification and fire_stale_reaped_notification functions (along with their tests) were deleted 
  from notify.rs, so if the stale watchdog was previously wired up anywhere, users have silently lost those notifications. Either the callers need to be added in this PR, or the notification deletions should be a separate follow-up once the new watchdog is integrated.

devinrosen (approved):

PR #1786, opened 2026-04-04
fix: resolve base branch from remote when not fetched locally

alabele:
## Summary
- **Auto-create local tracking branch**: When a Vantage ticket's `base_branch` exists on the remote but hasn't been fetched locally, `ensure_base_up_to_date` now creates a local tracking branch from `origin/{base}` instead of silently proceeding to a git failure
- **Clear error on missing branch**: Returns `"base branch 'X' not found locally or on remote 'origin'"` instead of letting `git branch` fail with a cryptic ref error
- **Prefix fallback**: New `resolve_and_update_base` tries `feat/` and `fix/` prefixes when the exact branch name isn't found (handles Vantage tickets that may omit the prefix)
- **Non-checkout fast-forward**: Updates non-checked-out base branches via local ref update (`git fetch .`) instead of checking out, avoiding unexpected working tree state changes

## Test plan
- [x] All 1200 conductor-core tests pass
- [x] Clippy clean (`-D warnings`)
- [x] `cargo fmt` clean
- [ ] Manual test: create worktree with `from_branch` pointing to a remote-only feature branch
- [ ] Manual test: create worktree with bare ticket name (no `feat/` prefix) when `feat/{name}` exists on remote


devinrosen (commented):
Only question:
Artifact cleanup silently dropped in CLI and TUI paths — This is the biggest behavioral regression in the PR. remove_artifacts was changed from a static method to an instance method so it can only be called with a WorktreeManager. Of the four callers of
  mark_worktrees_for_closed_tickets, only the MCP tool path passes Some(&worktree_mgr). The other three — conductor-cli/src/helpers.rs, conductor-tui/src/background.rs, and the sync_all path in tickets.rs — all pass None, meaning they silently skip artifact cleanup.     
  Before this PR, all paths cleaned up artifacts.

devinrosen (dismissed):
My question might be a non-issue. I think artifacts will still get cleaned up. Ship it. If it is an problem, we can make a follow up issue quickly.

alabele:
@devinrosen Good catch — you're right, this was a behavioral regression.

**The issue:** `remove_artifacts` was changed from a static method to an instance method, so `mark_worktrees_for_closed_tickets` gained an `Option<&WorktreeManager>` parameter. Three of the four callers (CLI helpers, TUI background, and the `sync_all` path in tickets.rs) passed `None`, silently skipping artifact cleanup. Only the MCP tool path passed `Some(&worktree_mgr)`.

**The fix (b1ac46d):** Reverted `remove_artifacts` back to a static method and call it as `WorktreeManager::remove_artifacts(...)` directly inside `mark_worktrees_for_closed_tickets`. This eliminates the `worktree_mgr` parameter entirely — all paths now clean up artifacts unconditionally, matching the pre-PR behavior.

devinrosen (approved):

PR #2491, opened 2026-04-23
feat(cli): add --bare mode support to reduce per-session startup tokens

viskystream:
## Summary

Each conductor-spawned Claude session pays a large fixed startup cost before any task-specific work: ~60k tokens of Claude Code auto-discovered context (CLAUDE.md, agent descriptions, skills catalog, plugin sync, auto-memory, hooks). Most of this is redundant for conductor, which already injects its own explicit session context via the prompt (ticket, worktree, plan steps, prior-run state).

Measured from one user's local session data (one week): ~600 conductor agent sessions. At ~60k startup overhead each, that's roughly **36M tokens per user per week** spent on Claude Code auto-discovery — unrelated to the actual work.

This PR adds `--bare` mode support at conductor's two Claude CLI call sites:

### 1. Plan generator (`conductor-cli/src/helpers.rs`) — unconditional

The plan call is a one-shot JSON prompt with no tool use and no project-context dependency. `--bare` is a drop-in win with no behavioral impact.

### 2. Agent runner (`conductor-cli/src/handlers/agent.rs`) — opt-in via new config flag

The agent loop benefits from `--bare` but enabling it has preconditions users should confirm, so it's gated behind a new `general.claude_bare_mode` flag (defaults to `false`):

```toml
[general]
claude_bare_mode = true
```

Preconditions users must satisfy before enabling:
- `ANTHROPIC_API_KEY` (or `apiKeyHelper` via `--settings`) must be set — `--bare` does not read OAuth/keychain auth
- User-configured Claude Code hooks will not fire in bare mode
- Any flow that relies on auto-discovered agents would need explicit `--agents` JSON injection (the common `general-purpose` / `Explore` paths are unaffected)

## Compatibility analysis (Site B)

| Feature `--bare` strips | Impact on conductor |
|---|---|
| CLAUDE.md auto-discovery | Safe — conductor already injects session context |
| Auto-memory | Safe — conductor has its own DB for run state |
| Plugin sync | Safe — conductor passes `--plugin-dir` explicitly |
| Background prefetches, LSP | Safe — not used by the one-shot agent loop |
| User hooks (`settings.json`) | Opt-in users accept this tradeoff |
| Agent catalog auto-load | Mostly safe — can use `--agents` for explicit injection when needed |
| Keychain auth | Opt-in users accept this precondition |

Conductor already has `ANTHROPIC_API_KEY` plumbing (`conductor-core/src/config.rs:569`, `workflow/api_call_executor.rs:35`), so the auth path exists.

## Estimated savings

Per-run savings (measured): ~25-40k tokens, with the upper end realized when the project has a dense CLAUDE.md and many project-scope agents/skills.

| Site | Calls/week (typical user) | Savings/call | Weekly savings |
|---|---|---|---|
| Plan generator | ~100-200 | ~25-40k | ~3-8M tokens |
| Agent runner (opt-in) | ~600 | ~25-40k | ~15-24M tokens |
| **Combined** | | | **~18-32M tokens/user/week** |

At scale across your user base, this is meaningful against Anthropic's rate limits and per-token costs.

## Architecture

Both changes preserve conductor's stateless-dispatch model. No behavioral changes when the flag is disabled (the default). The unconditional plan-generator change has no user-visible effect beyond lower token cost — plan JSON generation doesn't use any of the auto-discovered context.

## Not addressed here (future work)

The ~20k "Session Context" block conductor injects per session (measured: 82,912 bytes on the first user message) is separate overhead. If successive related task dispatches within the same worktree could share a session ID via `--resume` (rather than each starting fresh with re-injected state), Anthropic's prompt cache would make that header near-free on subsequent calls. This is a larger architectural change with its own tradeoffs (context accumulation, parallelism, error propagation) and is worth a separate discussion.

## Test plan

- [x] `cargo build -p conductor-cli -p conductor-core` — clean
- [x] `cargo test -p conductor-core --lib config::` — 220 passed including two new tests for the config flag
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --no-deps` — clean
- [ ] Manual verification: run `conductor agent run` with `claude_bare_mode = true` and `ANTHROPIC_API_KEY` set
- [ ] Manual verification: confirm plan generation still produces valid JSON output with `--bare`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## [Unreleased]
 
+### Added
+
+- Claude CLI `--bare` mode support for agent runs. Conductor now launches the
+  plan-generation call with `--bare` unconditionally (it's a one-shot JSON
+  prompt with no tool use or project-context needs). For the main agent runner,
+  bare mode is opt-in via a new `general.claude_bare_mode` config flag:
+
+  ```toml
+  [general]
+  claude_bare_mode = true
+  ```
+
+  When enabled, agent runs skip Claude Code's auto-discovered startup
+  (CLAUDE.md, agents, skills catalog, plugin sync, auto-memory, hooks),
+  saving roughly 25-40k tokens per run. Conductor's own session-context
+  injection is unaffected. Requires `ANTHROPIC_API_KEY` or `apiKeyHelper`
+  auth — keychain OAuth is not read in bare mode. Defaults to `false`.
+
 ### Deprecated
 
 - `[notifications.workflows]` — Use `[[notify.hooks]]` with `on` patterns instead.
--- a/conductor-cli/src/handlers/agent.rs
+++ b/conductor-cli/src/handlers/agent.rs
@@ -315,6 +315,13 @@ pub(crate) fn run_agent(
             .arg("stream-json")
             .arg("--verbose")
             .arg(effective_perm_mode.claude_permission_flag());
+        // Opt-in bare mode: skip Claude Code's auto-discovered context so conductor's
+        // explicit session-context injection isn't duplicated by the startup catalog.
+        // Saves ~25-40k tokens per run. Requires ANTHROPIC_API_KEY auth — see
+        // GeneralConfig::claude_bare_mode docs.
+        if config.general.claude_bare_mode {
+            cmd.arg("--bare");
+        }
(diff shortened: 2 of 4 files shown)


devinrosen (approved):
If you were able to run it and it works, send it!

devinrosen:
closing. need this base upstream to iterate.

PR #2496, opened 2026-04-23
feat(cli): add --bare mode support to reduce per-session startup tokens

devinrosen:
Each conductor-spawned Claude session pays a large fixed startup cost before any task-specific work: ~60k tokens of Claude Code auto-discovered context (CLAUDE.md, agent descriptions, skills catalog, plugin sync, auto-memory, hooks). Most of this is redundant for conductor, which injects its own explicit session context (ticket, worktree, plan steps, prior-run state).

At scale (measured ~600 agent sessions/week for one user), this overhead accounts for tens of millions of tokens per week in pure setup — unrelated to the actual work being done.

This change adds `--bare` mode support at two call sites:

1. Plan generator (conductor-cli/src/helpers.rs): unconditional. The plan call is a one-shot JSON prompt with no tool use and no project-context dependency, so --bare is a drop-in win with no behavioral impact.

2. Agent runner (conductor-cli/src/handlers/agent.rs): opt-in via the new `general.claude_bare_mode` config flag. The agent loop can benefit from --bare, but enabling it has preconditions users should confirm:
   - ANTHROPIC_API_KEY (or apiKeyHelper via --settings) must be set; --bare does not read OAuth/keychain auth.
   - User-configured Claude Code hooks will not fire in bare mode.
   - Any flow that relies on auto-discovered agents would need explicit --agents JSON injection.

Conductor already passes CLAUDE_CONFIG_DIR, --plugin-dir, --allowedTools, and other explicit context, so --bare fits cleanly with the existing stateless-dispatch architecture. No behavioral changes when the flag is disabled (the default).

Per-run savings (measured from local session data): ~25-40k tokens, with the upper end realized when the project under run has a dense CLAUDE.md and many project-scope agents/skills. At 600 sessions/week this yields ~15-24M tokens/week of pure overhead reduction for users who opt in, plus ~6M tokens/week from the unconditional plan-generator change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## [Unreleased]
 
+### Added
+
+- Claude CLI `--bare` mode support for agent runs. Conductor now launches the
+  plan-generation call with `--bare` unconditionally (it's a one-shot JSON
+  prompt with no tool use or project-context needs). For the main agent runner,
+  bare mode is opt-in via a new `general.claude_bare_mode` config flag:
+
+  ```toml
+  [general]
+  claude_bare_mode = true
+  ```
+
+  When enabled, agent runs skip Claude Code's auto-discovered startup
+  (CLAUDE.md, agents, skills catalog, plugin sync, auto-memory, hooks),
+  saving roughly 25-40k tokens per run. Conductor's own session-context
+  injection is unaffected. Requires `ANTHROPIC_API_KEY` or `apiKeyHelper`
+  auth — keychain OAuth is not read in bare mode. Defaults to `false`.
+
 ### Deprecated
 
 - `[notifications.workflows]` — Use `[[notify.hooks]]` with `on` patterns instead.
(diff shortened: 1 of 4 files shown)


devinrosen:
Going to close, not using ANTHROPIC_API_KEY.  Maybe eventual support.
⏺ There's your root cause — and commit 871a26a is not involved at all.                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                               
  You have claude_bare_mode = true in ~/.conductor/config.toml. The feat/claude-bare-mode PR being reviewed adds --bare to every Claude Code subprocess, but the code comment in that PR explicitly states:                                                                    
                                                                                                                                                                                                                                                                               
  ▎ Requires ANTHROPIC_API_KEY or apiKeyHelper auth — keychain OAuth is not compatible                                                                                                                                                                                         
                                                                                                                                                                                                                                                                               
  When Claude Code runs with --bare it skips the startup context including keychain auth, so even though you're logged in interactively, the subprocess doesn't see those credentials and reports "Not logged in". This affects every workflow run globally, which is why the  
  second unrelated qualify-ticket-comment run fails the same way.
                                                                                                                                                                                                                                                                               
  To fix it, either:                                                                                                                                                                                                                                                           
  1. Set ANTHROPIC_API_KEY in your environment, or                                                                                                                                                                                                                             
  2. Set claude_bare_mode = false in ~/.conductor/config.toml until you have the API key wired up

Issue #2569, opened 2026-04-24
Phase 3.2 — Replace resume_workflow_standalone internals with FlowEngine::resume()


devinrosen:
Part of Phase 3: Wire conductor-core through runkon-flow's FlowEngine. See [docs/runkon-flow-platform-spec.md](../blob/main/docs/runkon-flow-platform-spec.md#phase-3--wire-conductor-core-through-runkon-flows-flowengine-12-weeks).

## Context

Follows Phase 3.1 (merged, `02d50724`). With `execute_workflow_standalone` wired to `FlowEngine::run()`, the resume path also needs to delegate to the runkon-flow engine.

### 1. Add `FlowEngine::resume()` to `runkon-flow/src/flow_engine.rs`

```rust
pub fn resume(
    &self,
    def: &WorkflowDef,
    state: &mut ExecutionState,
) -> Result<WorkflowResult, EngineError>
```

Internally:
1. Call `state.persistence.get_steps(&state.workflow_run_id)` (already on `WorkflowPersistence` trait — no new trait methods required)
2. Build `skip_completed` / `step_map` from steps where `status == Completed`
3. Set `state.resume_ctx = Some(ResumeContext { skip_completed, step_map })`
4. Delegate to `self.run(def, state)`

The caller (conductor-core) is responsible for pre-modifying DB state before calling `resume()` — e.g. `reset_completed_steps` for `--restart`, `reset_steps_from_position` for `--from-step`. The engine reads whatever is in the DB, so no special arguments are needed for those cases.

### 2. Refactor `resume_workflow` in `conductor-core/src/workflow/engine.rs`

Keep all domain logic:
- `validate_resume_preconditions`, ephemeral run guard, orphaned step cleanup
- `reset_failed_steps` / `reset_completed_steps` / `reset_steps_from_position` for restart/from-step
- Worktree path resolution, run status update to Running

Replace the manual `ResumeContext` construction + `engine.run()` call (currently lines ~1496–1654) with:
- Build `ExecutionState` without setting `resume_ctx` (leave it `None`)
- Call `engine.resume(&rk_def, &mut rk_state)`

Also remove the stale NOTE comment in `resume_workflow_standalone` (line ~1340) which incorrectly claims the function uses the legacy `run_workflow_engine` path.

## Boundary

Gate resumability (precondition checks, `validate_resume_preconditions`) stays in conductor-core — conductor domain rules, not engine rules.

[...]

PR #2598, opened 2026-04-25
feat(#2569): phase 3.2 - replace resume internals with FlowEngine::resume()

devinrosen:
Adds FlowEngine::resume() to runkon-flow, which reads completed steps
from the post-reset DB state and builds ResumeContext internally before
delegating to run(). Refactors conductor-core's resume_workflow() to
remove the manual rk_step_map/rk_resume_ctx construction block, set
resume_ctx: None, and call engine.resume() instead of engine.run().
Removes a stale NOTE comment from resume_workflow_standalone.


devinrosen (commented):

**Reviewed by:** Error Handling, Test Coverage, Security, DRY & Abstraction, Architecture, Performance

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/engine.rs:235` — Layer inversion: `completed_keys_from_steps` lives in engine.rs but is consumed by the manager layer (`manager/recovery.rs:987` calls `crate::workflow::engine::completed_keys_from_steps`). The manager should not depend on the engine — the dependency arrow points the wrong way. Move `completed_keys_from_steps` to `conductor-core/src/workflow/manager/recovery.rs` or a sibling helpers file; it has no engine-specific dependencies so the move is mechanical.
</details>
<details>
<summary><b>DRY & Abstraction</b> — 3 issues</summary>

- **warning** `runkon-flow/src/flow_engine.rs:176` — Double pass over completed steps: `completed_keys_from_steps(&steps)` iterates and extracts keys for completed steps, then `step_map` immediately re-filters the same `steps` for `Completed` and re-computes the same key `(step_name.clone(), iteration as u32)`. The completed-status predicate and key expression are duplicated across two passes. Build `step_map` in a single pass and derive `skip_completed` from its keys to eliminate the duplication.
- **warning** `runkon-flow/src/flow_engine.rs:1759` — `CountAlpha` and `CountBeta` executor structs are defined identically (wrap `Arc<AtomicUsize>`, increment in `execute`, differ only in `name()`) in both `resume_skips_completed_steps` and `resume_empty_skip_set_runs_all` — four structurally identical struct definitions across two tests. Extract a single `CountingExecutor { name: &'static str, count: Arc<AtomicUsize> }` helper in the test module.
- **warning** `runkon-flow/src/flow_engine.rs:1797` — The `persistence.create_run(NewRun { workflow_name: "wf", worktree_id: None, ticket_id: None, ... })` block is copy-pasted verbatim in both `resume_skips_completed_steps` and `resume_empty_skip_set_runs_all`. An identical pattern already exists in `make_state_with_persistence` (line 1210). Extract a `make_test_run(p: &Arc<InMemoryWorkflowPersistence>) -> WorkflowRun` helper.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:175` — Error message in `resume()` omits the workflow run ID, making it harder to correlate failures in production logs. Suggestion: change the format string to include the run ID: `format!("resume: failed to load steps for run '{}': {e}", state.workflow_run_id)`
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:167` — FlowEngine::resume() has no test for the persistence.get_steps() error path. The map_err branch (EngineError::Workflow("resume: failed to load steps: …")) is reachable but untested. Add a test using a persistence stub whose get_steps() returns Err(_) and assert that engine.resume() propagates an EngineError::Workflow containing the 'resume: failed to load steps' prefix.
</details>

devinrosen (commented):

**Reviewed by:** Test Coverage, Security, Error Handling, DRY/Abstraction, Performance, Architecture

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/engine.rs:233` — Doc comment is now stale: it claims `completed_keys_from_steps` is 'Shared by `WorkflowManager::get_completed_step_keys` and `resume_workflow` so the key-building logic lives in one place.' After this PR, `get_completed_step_keys` in `recovery.rs` inlines the identical logic instead of calling this function, so the invariant ('logic lives in one place') is no longer true. Suggestion: Either (a) update the comment to state the function is now only used by the --from-step path in `resume_workflow`, or (b) have `recovery.rs:987` call `completed_keys_from_steps` again to restore single-source-of-truth.
</details>
<details>
<summary><b>DRY/Abstraction</b> — 1 issue</summary>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:987` — `get_completed_step_keys` now inlines the filter+map+collect logic instead of delegating to `completed_keys_from_steps` in engine.rs. The shared helper exists precisely for this method — its docstring still says "Shared by `WorkflowManager::get_completed_step_keys`" — but that relationship was severed here, creating a duplicate copy of the same logic. Suggestion: Restore the one-line delegation: `Ok(crate::workflow::engine::completed_keys_from_steps(&steps))`. If the intent was to decouple from engine internals, update the helper's docstring to drop the stale reference.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:167` — The doc comment requires `state.resume_ctx` to be `None` on entry, but there is no runtime enforcement. When `skip_completed` is empty the conditional branch (line 187) is skipped, leaving any pre-existing `resume_ctx` intact and passing it to `run()`. The current caller satisfies the contract, but the invariant is invisible to future callers. Suggestion: Add `debug_assert!(state.resume_ctx.is_none(), "resume() requires resume_ctx to be None on entry");` as the first line of the method body so tests catch contract violations early.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:186` — `step_map` is iterated a second time via `.keys().cloned().collect()` to build `skip_completed`, producing N redundant String clones. Both collections derive from the same source and can be built in a single pass. Suggestion: Use a fold or manual loop to unzip both collections simultaneously — this halves the iteration cost and eliminates N redundant String heap allocations (one per completed step).
</details>



- [#2601 — Identical completed_keys_from_steps duplicated across conductor-core and runkon-flow](https://github.com/devinrosen/conductor-ai/issues/2601) — `conductor-core/src/workflow/engine.rs:235` (warning)

devinrosen (commented):

**Reviewed by:** Test Coverage, Security, Dry Abstraction, Error Handling, Architecture, Performance

### Blocking findings

<details>
<summary><b>Architecture</b> — 2 issues</summary>

- **warning** `runkon-flow/src/flow_engine.rs:29` — `EngineBundle` is a backward-compatibility shim with zero consumers outside this crate. A grep across the repo shows it is only referenced in `runkon-flow/src/flow_engine.rs` and `runkon-flow/src/lib.rs`. The project convention explicitly states: "Avoid backwards-compatibility hacks like re-exporting types." Keeping a dead public re-export bloats the crate's API surface and misleads callers into thinking the type is still meaningful. Suggestion: Remove the `EngineBundle` struct from `flow_engine.rs` and drop its `pub use` line from `lib.rs`.
- **warning** `conductor-core/src/workflow/engine.rs:1460` — The `skip_completed` `HashSet<StepKey>` is fully constructed (including the `from_step` pruning walk) but is then only consumed for a single `skip_count` log line (line 1494). The actual skip-set that drives execution is independently rebuilt by `FlowEngine::resume()` from the post-reset DB state. The DB-mutating side-effects of this block (`reset_failed_steps`, `reset_completed_steps`, `reset_steps_from_position`) are still required and correct — only the HashSet construction is now redundant. This creates a misleading "compute and discard" pattern. Suggestion: Replace the HashSet construction with a plain integer counter (or a direct count query) for logging only; keep all DB mutation calls in place.
</details>
<details>
<summary><b>Dry Abstraction</b> — 2 issues</summary>

- **warning** `runkon-flow/src/flow_engine.rs:1940` — `FailingSteps` reimplements all 14 `WorkflowPersistence` methods as `unimplemented!()` just to fail on `get_steps`. This diverges from the established fault-injection pattern already in `InMemoryWorkflowPersistence` (`fail_get_fan_out_items: AtomicBool` + `set_fail_get_fan_out_items(bool)`). Every future method added to the `WorkflowPersistence` trait will require a corresponding `unimplemented!()` stub here, creating silent maintenance debt across ~90 lines. Suggestion: Add `fail_get_steps: AtomicBool` and `set_fail_get_steps(bool)` to `InMemoryWorkflowPersistence` (mirroring `fail_get_fan_out_items`), check the flag in its `get_steps` impl, and replace the entire `FailingSteps` struct + its trait impl with 3–4 lines using `persistence.set_fail_get_steps(true)`.
- **warning** `runkon-flow/src/flow_engine.rs:1843` — The 14-line block that constructs alpha/beta `CountingExecutor` instances and wires them into `state.action_registry` is copy-pasted verbatim between `resume_skips_completed_steps` (lines 1843–1861) and `resume_empty_skip_set_runs_all` (lines 1892–1911). The existing `make_bare_state` helper establishes that this module extracts repeated state construction. Suggestion: Add a `make_counting_state(persistence: Arc<InMemoryWorkflowPersistence>, run_id: &str) -> (Arc<AtomicUsize>, Arc<AtomicUsize>, ExecutionState)` helper alongside `make_bare_state` and `make_test_run`.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1631` — `engine.resume()` error is wrapped as `ConductorError::Workflow(e.to_string())` with no added context. Three lines above, the engine-build error includes the workflow name ("failed to build engine for '{}': {e}"), but the resume error drops that context. The inner `EngineError` does embed the run ID, so this is debuggable — but the inconsistency means errors from the resume path look different from build-path errors in logs. Suggestion: `engine.resume(&rk_def, &mut rk_state).map_err(|e| ConductorError::Workflow(format!("workflow '{}' resume failed: {e}", workflow.name)))?;`
</details>



- [#2602 — ~150 lines of `ExecutionState` construction duplicated between `execute_workflow_standalone` and `resume_workflow`](https://github.com/devinrosen/conductor-ai/issues/2602) — `conductor-core/src/workflow/engine.rs:1113` (warning)

devinrosen (commented):

**Reviewed by:** Error Handling, Security, Test Coverage, Architecture, Performance, DRY Abstraction

### Blocking findings

<details>
<summary><b>Architecture</b> — 3 issues</summary>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:984` — `WorkflowManager::get_completed_step_keys` doc comment says 'Used to build the skip set for resume' but skip-set construction now lives exclusively in `FlowEngine::resume()`. This method has no production callers — only tests — yet its documentation implies conductor-core still participates in skip-set construction. The companion `completed_keys_from_steps` (engine.rs:233) carries the same stale claim. Update the doc comments or remove the methods if the tests can be re-expressed through the runkon-flow API.
- **warning** `runkon-flow/src/flow_engine.rs:159` — `debug_assert!(state.resume_ctx.is_none(), ...)` enforces a public cross-crate API precondition that is compiled away in `--release` builds. If a caller pre-seeds `resume_ctx: Some(...)` before calling `resume()`, the pre-seeded context is silently used in production with no error, violating the method's ownership of skip-set construction. Replace with an explicit guard returning `Err(EngineError::Workflow(...))` to enforce the precondition in all build profiles.
- **warning** `conductor-core/src/workflow/engine.rs:1391` — The definition snapshot is deserialized twice in `resume_workflow`: once into conductor-core's `crate::workflow_dsl::WorkflowDef` (line 1391) and again into `runkon_flow::dsl::WorkflowDef` (line 1512). No shared version guard exists. If the two DSL types silently diverge (a field renamed, a new required field added), the second deserialization fails at runtime with an opaque JSON error and no indication of which type diverged. At minimum, add a `definition_version` field to snapshots; a stronger fix is to make conductor-core's `WorkflowDef` a re-export of the runkon-flow type.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1512` — `resume_workflow` deserializes `snapshot` twice: once at line 1391 into `workflow: WorkflowDef` (conductor-core type) and once here into `rk_def` (runkon-flow type). The first parse exists only to provide `workflow.name`, but `wf_run.workflow_name` is already loaded and can be used instead for logging, the schema resolver closure, the `ExecutionState` `workflow_name` field, and error messages. `execute_workflow_standalone` in the same file already avoids this pattern. Remove the first deserialization and replace `workflow.name` references with `wf_run.workflow_name`.
- **warning** `runkon-flow/src/persistence_memory.rs:66` — The `fail_get_steps` fault-injection added here is structurally identical to the existing `fail_get_fan_out_items` pattern: an `AtomicBool` field, a `set_fail_X(&self, bool)` setter, and a guard at the top of the method. Two instances is acceptable, but the boilerplate will accumulate if more methods need fault injection. No action required now; consider a macro if a third injection point is added.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:159` — `debug_assert!` for the `resume_ctx.is_none()` precondition is compiled away in release builds. If a caller passes state with a pre-populated `resume_ctx` and the DB returns zero completed steps, the condition on line 181 (`if !skip_completed.is_empty()`) is false, so the old context is never overwritten — silently causing steps that should execute to be skipped instead. Promote to a hard `assert!` or return an `Err` early to enforce the invariant in production.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1364` — `resume_workflow()` issues two full reads of `workflow_run_steps` for the same `run_id`. The first read (`all_steps`, line 1364) is used only for `from_step` validation and position lookup. Then `engine.resume()` opens a new connection and calls `persistence.get_steps()` again to rebuild the skip set. In the common resume case (no restart, no `--from-step`) no completed steps are reset between the two reads, so the second query returns identical data. Accept `all_steps` as an optional parameter into `FlowEngine::resume()` to skip the redundant query.
- **warning** `runkon-flow/src/flow_engine.rs:172` — The fold builds both `step_map: HashMap<StepKey, _>` and `skip_completed: HashSet<StepKey>` in lockstep, but `skip_completed` is always exactly `step_map.keys().collect()`. Every completed step pays an extra `String` clone for set insertion and the `HashSet` occupies a separate allocation. Both call-sites (`should_skip()` at engine.rs:710 and `find_max_completed_while_iteration()` at helpers.rs:249) use only `contains()`, which is equivalent to `step_map.contains_key()`. Remove `skip_completed` from `ResumeContext` and replace all uses with `step_map` lookups.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:149` — The `debug_assert!(state.resume_ctx.is_none())` precondition in `FlowEngine::resume()` has no test that verifies a violation causes a panic. The call site in `engine.rs` satisfies the invariant statically, but a future caller could accidentally pass a pre-populated `resume_ctx` and the assertion would silently pass in release builds. Add a `#[test] #[should_panic]` case or convert to a hard `assert!`.
</details>

devinrosen (commented):

**Reviewed by:** Test Coverage, Security, Error Handling, Performance, Architecture, DRY Abstraction

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:987` — WorkflowManager::get_completed_step_keys calls crate::workflow::engine::completed_keys_from_steps — a manager method depending on an engine-layer free function. The function is a two-line pure data transformation (filter + map on WorkflowRunStep) with no engine-specific logic. Now that FlowEngine::resume() builds the skip set internally, this method has no remaining production callers (only tests), so the engine coupling exists purely to support test assertions. Inline the transformation directly in get_completed_step_keys or move it to a #[cfg(test)] block.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:681` — Silent discard of `open_database` error in best-effort notification path inside a spawned thread. If the DB open fails, the notification is skipped with no log entry, leaving no trace to diagnose missing stuck-run notifications. Add an `else` branch: `} else { tracing::warn!(run_id = %run_id_clone, "reap_heartbeat_stuck_runs: could not open DB to fire stuck-run notification"); }` — preserves the fire-and-forget intent while leaving a trace when the DB is unavailable.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:166` — FlowEngine::resume() issues a get_steps() DB query even though the calling code in resume_workflow (engine.rs:1364) already fetched the same steps table. The double-fetch is intentional (resume() needs post-reset state), but it contradicts the comment on line 1363 ('Load all steps once — avoids N+1') and adds an extra SQLite SELECT per resume invocation. Either update the comment on engine.rs:1364 to acknowledge the post-reset re-fetch, or accept a pre-fetched step slice in resume() and let the caller decide when to re-query after all resets complete.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1619` — `resume_workflow()` has a material behavioral change — it now passes `resume_ctx: None` and calls `engine.resume()` (which reads DB post-reset) instead of building a `skip_completed` HashSet up-front and calling `engine.run()` — but no new test in the diff validates the end-to-end integration path from `resume_workflow()` through `FlowEngine::resume()`. Add a conductor-core integration test (or extend an existing one) that invokes `resume_workflow()` or `resume_workflow_standalone()` against an in-memory or test DB with a partially-completed run, and asserts that already-completed steps are skipped and remaining steps execute.
</details>



- [#2603 — WorkflowManager recovery methods spawn engine execution threads — bidirectional manager↔engine coupling](https://github.com/devinrosen/conductor-ai/issues/2603) — `conductor-core/src/workflow/manager/recovery.rs:572` (warning)
- [#2604 — call_node test helper defined three times identically](https://github.com/devinrosen/conductor-ai/issues/2604) — `runkon-flow/src/helpers.rs:378` (warning)
- [#2605 — ExecutionState construction duplicated between execute and resume paths](https://github.com/devinrosen/conductor-ai/issues/2605) — `conductor-core/src/workflow/engine.rs:1564` (warning)

devinrosen (commented):

**Reviewed by:** Security, Test Coverage, Dry Abstraction, Performance, Error Handling, Architecture

### Blocking findings

<details>
<summary><b>Architecture</b> — 4 issues</summary>

- **warning** `runkon-flow/src/engine.rs:22` — `ENGINE_INJECTED_KEYS` in runkon-flow contains conductor-core domain keys (`ticket_source_id`, `ticket_raw_json`, `repo_path`, etc.). runkon-flow is a general-purpose workflow engine crate — it should not encode conductor-specific domain concepts. The canonical declaration belongs in `conductor-core` (already at `conductor-core/src/workflow/engine.rs:27`). Suggestion: Remove `ENGINE_INJECTED_KEYS` from `runkon-flow/src/engine.rs` and keep conductor-core's copy as the single source of truth.
- **warning** `runkon-flow/src/engine.rs:215` — `completed_keys_from_steps` is duplicated verbatim in `conductor-core/src/workflow/engine.rs:235`. `recovery.rs:987` calls the conductor-core copy; any future divergence (e.g. adding `TimedOut` to the filter) will silently affect only one path. Suggestion: Keep the function in `runkon-flow/src/engine.rs` as canonical and import it in conductor-core via `use runkon_flow::engine::completed_keys_from_steps`, deleting the local copy.
- **warning** `conductor-core/src/workflow/flow_engine.rs:13` — `conductor-core::workflow::flow_engine::FlowEngineBuilder` builds an `ActionRegistry`, while `runkon-flow::flow_engine::FlowEngineBuilder` builds a `FlowEngine`. Same name, fundamentally different return types. As the phase 3.x migration advances, this collision will create ambiguous paths and confusing compiler errors. Suggestion: Rename the conductor-core builder to `ActionRegistryBuilder` — only `engine.rs:85` and its tests reference it.
- **warning** `runkon-flow/src/flow_engine.rs:104` — `FlowEngine::run()` accepts a `runkon_flow::engine::ExecutionState` carrying runkon-flow's `ActionRegistry`. conductor-core's execution path builds its own `conductor-core::workflow::action_executor::ActionRegistry` — a separate, structurally identical but type-incompatible type. This means `FlowEngine::run()` cannot be called from conductor-core's resume path without an adapting bridge, blocking the stated phase 3.2 migration goal. Suggestion: Phase 4/3.3 should either re-export `runkon_flow::traits::action_executor::*` directly or add a `From`/`Into` adapter between the two `ActionRegistry` types.
</details>
<details>
<summary><b>Dry Abstraction</b> — 2 issues</summary>

- **warning** `runkon-flow/src/flow_engine.rs:1569` — Inline NewRun literal bypasses the `make_test_run()` helper defined at line 646 for the same purpose. The same 12-field default literal appears at lines 1569, 1684, 1775, and 1957 while the two resume tests at lines 1825 and 1881 correctly use `make_test_run()`. Suggestion: Replace the four inline `persistence.create_run(NewRun { workflow_name: "wf", ... })` blocks with `let run = make_test_run(&persistence);`.
- **warning** `runkon-flow/src/flow_engine.rs:1253` — `VecSink`, `ForwardSink`, `call_node`, and `make_def` are defined inline in this file's `#[cfg(test)]` block but are structurally identical to the versions in `tests/common/mod.rs`. The runkon-flow crate already has a `src/test_helpers.rs` module gated `#[cfg(test)]` for exactly this purpose. Suggestion: Move these helpers into `src/test_helpers.rs`; integration test versions in `tests/common/mod.rs` remain as-is due to Rust's test module boundary.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1425` — `expect("guarded above")` won't be useful in a crash log — the guard it references is ~40 lines earlier and a reader hitting this panic in production logs would have no idea which invariant fired. Suggestion: Use a self-contained message: `.expect("ticket_id is Some when worktree_id and repo_id are both None — enforced by the ephemeral run guard above")`.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:263` — Unnecessary intermediate `Vec<String>` allocation: `parent_ids` is built by cloning all `parent_run_id` strings only to immediately borrow from it to build `id_refs: Vec<&str>`. This allocates N strings that live only long enough to create N borrows. Suggestion: Drop `parent_ids` entirely and build `id_refs` directly from `waiting_runs` via `.iter().map(|(_, id)| id.as_str()).collect()`.
- **warning** `runkon-flow/src/helpers.rs:116` — O(N_keys × N_completed_iters) transient String allocations in `find_max_completed_while_iteration`: `k.clone()` inside the inner `all()` creates a temporary owned String per body key per completed while-loop iteration, discarded immediately after the HashSet lookup. Suggestion: Build a per-iteration set of completed step names (`HashSet<&str>`) before the `all()` check to avoid per-key clones, or model the skip set as `HashMap<String, HashSet<u32>>`.
- **warning** `conductor-core/src/workflow/manager/recovery.rs:284` — N+1 DB query pattern in `reap_orphaned_workflow_runs`: `self.find_waiting_gate(&run_id)` is called once per entry in the `waiting_runs` loop, producing one extra SELECT per orphaned waiting run after the initial batch fetch. Suggestion: Fetch the waiting gate step alongside the initial query with a LEFT JOIN on `workflow_run_steps` where `status = 'waiting'`, collapsing N+1 queries into one.
</details>
<details>

- **warning** `runkon-flow/src/helpers.rs:248` — `find_max_completed_while_iteration` was changed from `skip_set.contains(...)` (HashSet) to `step_map.contains_key(...)` (HashMap), but no test in the diff exercises this function with the new `ResumeContext.step_map` field. A while-loop resume scenario — where the engine must compute the highest fully-completed iteration before re-entering the loop — is a distinct code path not exercised by any of the four new `FlowEngine::resume()` unit tests. Suggestion: Add a `FlowEngine::resume()` test that uses a `while` node: pre-seed two complete iterations of the while body in `InMemoryWorkflowPersistence`, call `engine.resume()`, and assert that the loop body executors are only called for the third (incomplete) iteration.
</details>



- [#2606 — Duplicated CAS-flip + spawn pattern in two watchdog functions](https://github.com/devinrosen/conductor-ai/issues/2606) — `conductor-core/src/workflow/manager/recovery.rs:506` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Test Coverage, Security, DRY/Abstraction, Performance, Error Handling, Architecture

### Blocking findings

<details>
<summary><b>Architecture</b> — 2 issues</summary>

- **warning** `runkon-flow/src/engine.rs:48` — ResumeContext omits child-run metrics, creating a silent cost divergence. conductor-core's ResumeContext carries `child_runs: HashMap<String, AgentRun>` and `restore_completed_step` accumulates cost_usd, num_turns, duration_ms, and all token fields from those pre-loaded runs. runkon-flow's ResumeContext has no such field, so its `restore_completed_step` (lines 735-783) produces zero accumulated costs for restored steps. When conductor-core migrates to FlowEngine::run() as the production resume path, total_cost/total_turns/token counts for resumed runs will silently undercount all costs from already-completed steps. Suggestion: Store per-step cost fields directly on WorkflowRunStep and accumulate them in `restore_completed_step` via a WorkflowPersistence query.
- **warning** `conductor-core/src/workflow/manager/recovery.rs:576` — WorkflowManager (data-access layer) spawns execution engine threads. `auto_resume_stuck_workflows`, `reap_heartbeat_stuck_runs`, and `watchdog_needs_resume_workflows` all call `std::thread::spawn` with `resume_workflow_standalone`. Per the manager pattern (CLAUDE.md), managers own CRUD/query logic on their domain entity; orchestration-level work — spawning background resumption threads, firing notifications — belongs in the calling layer. This creates a bidirectional dependency: the engine calls WorkflowManager, and WorkflowManager calls back into the engine. Suggestion: Have these methods return `Vec<String>` of run IDs that need resumption and move the thread-spawn + resume_workflow_standalone calls to the TUI or CLI watchdog.
</details>
<details>
<summary><b>DRY/Abstraction</b> — 1 issue</summary>

- **warning** `conductor-core/src/workflow/tests/mod.rs:27` — The `completed_keys_from_steps` wrapper converts core steps to runkon-flow types just to delegate to `runkon_flow::engine::completed_keys_from_steps`, but `recovery.rs:get_completed_step_keys` (line 989) already inlines the same 3-line filter+map+collect pattern directly on core types. Two implementations now exist and the comment's claim of 'avoiding duplication' doesn't hold. Suggestion: Inline the same pattern as `get_completed_step_keys`: `steps.iter().filter(|s| s.status == WorkflowStepStatus::Completed).map(|s| (s.step_name.clone(), s.iteration as u32)).collect()`. Remove the `step_to_rk` conversion round-trip and the delegation to `runkon_flow::engine::completed_keys_from_steps`.
</details>
<details>

- **warning** `runkon-flow/src/prompt_builder.rs:112` — `.unwrap_or_default()` returns `""` (empty string) when serialization fails for a non-empty contexts vec. Templates referencing `{{prior_contexts}}` receive an empty string rather than a valid JSON array, and no warning is emitted. Suggestion: Replace with `serialize_or_empty_array(&state.contexts, "build_variable_map:prior_contexts")` from `crate::helpers`, which falls back to `"[]"` and logs a warning on failure.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/queries.rs:793` — `prepare_cached` used with dynamically-generated SQL whose shape changes with `run_ids.len()`. Every distinct batch size produces a different SQL string and therefore a distinct cache entry. Over the lifetime of the process the statement cache accumulates one compiled statement per unique batch size seen, growing without bound. Suggestion: Use `prepare()` instead of `prepare_cached()` for IN-list queries with variable placeholder counts. Caching is only effective when the SQL string is stable across calls; here it is not.
- **warning** `runkon-flow/src/helpers.rs:263` — `completed_by_iter` is built by scanning every key in `step_map` — the entire run's step history — even though only body-key entries are relevant for the while-loop check. For a resumed run with N completed steps and a while-loop whose body has k steps, when N >> k × completed_iterations (e.g. a large workflow with hundreds of completed non-while steps and a small, shallow while-loop), the new code does strictly more work than the old O(k × completed_iterations) approach. Suggestion: Filter `step_map` keys to only those whose name appears in `body_keys` before building `completed_by_iter` using a `HashSet<&str>` for O(1) membership checks.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/queries.rs:289` — `find_waiting_gates_for_runs` is a new public method with non-trivial logic (dynamic SQL placeholder building, batch query, position-based deduplication via `or_insert` on DESC-ordered results) but has no unit test in the diff. A bug in the dedup logic or the SQL would silently return the wrong gate step for affected runs, breaking gate-timeout reaping. Suggestion: Add a unit test that (1) verifies an empty `run_ids` slice returns an empty map immediately; (2) inserts two runs each with multiple gate steps at different positions and asserts the returned map contains only the highest-position step per run; (3) verifies runs with no waiting gate are absent from the map.
</details>



- [#2607 — Pre-existing: CAS run-status flip pattern duplicated across 3 watchdog functions](https://github.com/devinrosen/conductor-ai/issues/2607) — `conductor-core/src/workflow/manager/recovery.rs:541` (warning)
- [#2608 — Pre-existing: resume thread spawn pattern duplicated across 3 watchdog functions](https://github.com/devinrosen/conductor-ai/issues/2608) — `conductor-core/src/workflow/manager/recovery.rs:571` (warning)

devinrosen (commented):

**Reviewed by:** Security, Performance, DRY Abstraction, Test Coverage, Architecture, Error Handling

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/engine.rs:1568` — `resume_workflow` accepts `input.conn` (`&Connection`) for the pre-execution phase (step resets, status updates, orphan cleanup) but opens a second independent SQLite connection (`raw_conn`) for the FlowEngine execution phase. Correctness relies on SQLite WAL auto-commit visibility across connections rather than an explicit API contract. By contrast, `execute_workflow_standalone` is fully self-contained — it opens exactly one connection. The asymmetry creates a subtle ordering requirement: `input.conn` mutations must auto-commit before `raw_conn` is opened. Suggestion: Make `resume_workflow` fully self-contained by removing `input.conn` from `WorkflowResumeInput` and opening one connection for both phases, matching `execute_workflow_standalone`'s pattern.
</details>
<details>
<summary><b>DRY Abstraction</b> — 3 issues</summary>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:542` — Identical CAS-flip SQL duplicated in `claim_stuck_workflows` (lines 542–548) and `claim_heartbeat_stuck_runs` (lines 620–626). Both execute `UPDATE workflow_runs SET status = 'failed', error = 'Orphaned: executor died between steps — auto-resumed by watchdog' WHERE id = :id AND status = 'running'` verbatim. If the error string changes or a new column needs updating, both sites must be kept in sync. Suggestion: Extract a private helper `fn cas_flip_run_to_failed(&self, run_id: &str, from_status: &str, error_msg: &str) -> Result<bool>` and call it from both claim functions.
- **warning** `conductor-core/src/workflow/manager/recovery.rs:587` — `claim_heartbeat_stuck_runs` inlines the orphaned-run detection WHERE clause (lines 589–600) instead of reusing `detect_stuck_workflow_run_ids`. The WHERE clause is word-for-word identical; only the SELECT columns differ. `claim_stuck_workflows` correctly delegates to `detect_stuck_workflow_run_ids`, creating an inconsistency: a filter change must be applied in two places. Suggestion: Add `detect_stuck_workflow_run_ids_with_meta(threshold_secs: i64) -> Result<Vec<(String, String, Option<String>)>>` that selects the extra columns, and replace the inline query in `claim_heartbeat_stuck_runs` with a call to it.
- **warning** `runkon-flow/src/prompt_builder.rs:8` — `ENGINE_INJECTED_KEYS` is defined identically in both `runkon-flow/src/prompt_builder.rs` (line 8, private `const`) and `conductor-core/src/workflow/engine.rs` (line 27, `pub(crate) const`). Both list the same 11 keys. Adding a new injected variable requires updating both files, with no compile-time guard against divergence. Suggestion: Expose the constant from `runkon-flow` as `pub const ENGINE_INJECTED_KEYS: &[&str]` and have `conductor-core` import and re-export it.
</details>
<details>

- **warning** `runkon-flow/src/persistence_memory.rs:403` — `approve_gate` silently swallows a serde serialization failure with `unwrap_or_default()` for `gate_selections`, storing an empty string without any log entry. The conductor-core equivalent at `persistence_memory.rs:361` logs `tracing::warn!("approve_gate: failed to serialize selections: {e}")` before falling back. While `Vec<String>` serialization is practically infallible, the silent discard diverges from the established logging convention. Suggestion: Replace `serde_json::to_string(s).unwrap_or_default()` with `serde_json::to_string(s).unwrap_or_else(|e| { tracing::warn!(…); String::new() })` to match the conductor-core pattern.
</details>
<details>
<summary><b>Performance</b> — 4 issues</summary>

- **warning** `conductor-tui/src/background.rs:698` — N+1 query per active run: `wf_mgr.get_workflow_steps(&run.id)` is called inside the loop over `active_runs` with no caching. With 10 concurrent runs this is 10 sequential DB queries per poll tick. Suggestion: Batch-fetch all step rows for the relevant run IDs in a single query before the loop (e.g. `get_workflow_steps_for_runs(&[&run.id, …])`) or extend the existing `step_history_cache` keyed on `run.id` using the `Entry` API.
- **warning** `conductor-tui/src/background.rs:674` — `get_completed_run_durations` is called in the fallback branch inside the loop but is not cached. For every active run whose workflow has no step history, this fires an additional uncached DB query per poll tick. Suggestion: Add a second `run_history_cache: HashMap<String, Vec<i64>>` keyed on `workflow_name` (mirroring the `step_history_cache` pattern above) so the fallback query is also deduplicated per workflow name per tick.
- **warning** `conductor-core/src/workflow/persistence_memory.rs:256` — O(n) linear scan on `fan_out_items.values()` for idempotency check in `insert_fan_out_item`. For a foreach step dispatching N items this is O(n²) total. The `runkon-flow` copy of this struct already has a `fan_out_index: HashMap<(String, String), String>` secondary index that makes this O(1). Suggestion: Mirror the `runkon-flow/src/persistence_memory.rs` pattern: add `fan_out_index: HashMap<(String, String), String>` to `InMemoryStore` and replace the `.values().find(…)` scan with a single `store.fan_out_index.get(&(step_run_id, item_id))` lookup.
- **warning** `conductor-core/src/workflow/manager/recovery.rs:218` — One `update_step_status_full` call per stuck step — N sequential UPDATE round trips when multiple steps are stuck simultaneously. Each call issues at least one SQL statement. Suggestion: Consider a single `UPDATE workflow_run_steps SET status = … WHERE id IN (…)` for the common case where all stuck steps share the same target status (Completed or Failed), falling back to the per-row path only for the mixed-status minority.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:62` — `spawn_heartbeat_resume` is a new `pub fn` with a non-trivial failure path (DB open + `fire_heartbeat_stuck_failed_notification` call, plus a fallback warn when DB open itself fails) and has no test in the diff. If the notification path regresses it will fail silently. Suggestion: Add a unit test using a `FailingPersistence` stub (similar to `resume_propagates_get_steps_error`) that verifies the error branch executes without panicking, or an integration test that asserts the notification is fired when `resume_workflow_standalone` returns an error.
</details>

devinrosen (commented):

**Reviewed by:** DB Migrations, Security, DRY/Abstraction, Test Coverage, Error Handling, Performance, Architecture

### Blocking findings

<details>
<summary><b>Architecture</b> — 2 issues</summary>

- **critical** `conductor-core/src/workflow/runkon_bridge.rs:234` — RkActionExecutorAdapter::execute() acquires the DB mutex, creates an AgentRun row via AgentManager, and updates workflow_run_steps.child_run_id via WorkflowManager — all from inside an ActionExecutor trait implementation. Executors sit at the execution layer; they should be pure input→output transformers. Creating and linking persistence rows is an orchestration concern that belongs one level up: in FlowEngine's step dispatch or conductor-core's step lifecycle, not inside the executor adapter. Suggestion: Move create_child_run() and update_step_child_run_id() out of the executor. One clean option: add an optional child_run_id field to ActionOutput (already present in the runkon-flow type, currently set to None at line 175) and have the engine consume it post-execution to write the DB link. This removes the need for conn and the AgentManager/WorkflowManager imports inside the adapter entirely.
- **warning** `conductor-desktop/src/main.rs:77` — conductor-desktop calls claim_stuck_workflows() while conductor-web calls claim_heartbeat_stuck_runs() — a different method with different semantics. Desktop is also missing reap_finalization_stuck_workflow_runs and reap_stale_workflow_runs, both present in the web startup path. Both binaries are thin consumers of conductor-core; their startup cleanup sequences should be identical. Suggestion: Add a WorkflowManager::run_startup_reaps(&config) method (or similar) in conductor-core that encapsulates the full ordered set of cleanup calls. Both conductor-web and conductor-desktop call only that one method, eliminating the drift.
</details>
<details>
<summary><b>DRY/Abstraction</b> — 4 issues</summary>

- **warning** `runkon-flow/tests/common/mod.rs:95` — VecSink and ForwardSink (lines 95–128) are copy-pasted verbatim from runkon-flow/src/test_helpers.rs (lines 41–71). The only differences are field visibility (pub vs private) and doc-comment wording. tests/common/mod.rs can re-export both types from test_helpers instead of re-implementing them. Suggestion: Remove the VecSink/ForwardSink definitions from tests/common/mod.rs and re-export them: `pub use runkon_flow::test_helpers::{VecSink, ForwardSink};`. Adjust test_helpers.rs to make the `events` field pub if needed.
- **warning** `conductor-core/src/workflow/rk_types.rs:30` — run_status_to_core (lines 30–45) and run_status_to_rk (lines 47–62) are mirror match-tables over 8 identical variant names. Same pattern repeats for step_status_to_core/step_status_to_rk (lines 64–94). Four parallel match tables to maintain when a variant is added. Suggestion: Replace each function pair with From impls: `impl From<runkon_flow::status::WorkflowRunStatus> for crate::workflow::status::WorkflowRunStatus` and the reverse. Each arm stays a trivial rename, but callers use .into() and new variants produce compile errors in one place instead of two.
- **warning** `conductor-core/src/workflow/rk_types.rs:289` — core_workflow_result_to_rk (lines 289–305) and rk_workflow_result_to_core (lines 307–323) both do field-by-field copies with no type conversion — every field has the same name and type in both structs. These are a symmetric pair that would be cleaner and safer (compiler catches missed fields) as From impls. Suggestion: Replace with `impl From<CoreResult> for RkResult` and `impl From<RkResult> for CoreResult` in rk_types.rs.
- **warning** `conductor-core/src/workflow/rk_types.rs:275` — fan_out_item_to_rk (lines 275–287) copies every field of CoreFanOutItemRow to RkFanOutItemRow with no conversion. All 9 fields are direct assignments. A From impl would be more idiomatic and would generate a compile error if either struct gains a field. Suggestion: Replace with `impl From<CoreFanOutItemRow> for RkFanOutItemRow` and delete fan_out_item_to_rk. Update the single call site in runkon_bridge.rs.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1387` — `Err(_)` discards the database-open error in the heartbeat-resume notification path, making the warning message unactionable — you can see it fired but not why the DB could not be opened (missing file, permission denied, WAL corruption, etc.). Suggestion: Change `Err(_) =>` to `Err(e) =>` and include `{e}` in the tracing::warn! message.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:284` — N+1 query pattern: `find_waiting_gate` issues a separate DB SELECT per waiting run inside the loop. With N runs in `waiting` status, `reap_orphaned_workflow_runs` executes N+1 queries (1 batch fetch of runs + 1 per run for gate step). Under load this scales linearly with waiting-run count. Suggestion: Batch-fetch gate steps for all waiting run IDs before the loop using a single IN-clause query, then look up from a local `HashMap<run_id, Option<WorkflowRunStep>>` inside the loop.
- **warning** `conductor-core/src/workflow/engine.rs:321` — O(n²) deduplication: `all_plugin_dirs.contains(&dir)` is O(m) and is called once per item in `collect_all_plugin_dirs()`, making the overall dedup O(k×m). Negligible for typical small plugin-dir counts but a HashSet would be strictly better. Suggestion: Use a `HashSet<String>` seeded from `input.extra_plugin_dirs` as a seen-set, then push to `all_plugin_dirs` only on a miss.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:131` — `spawn_workflow_resume` is a new public function with no test. It spawns a background thread, constructs `WorkflowResumeStandalone`, calls `resume_workflow_standalone`, and logs on error. The sibling function `spawn_heartbeat_resume` has a test (`test_spawn_heartbeat_resume_handles_failed_resume_gracefully`); this one does not. Suggestion: Add a test similar to `test_spawn_heartbeat_resume_handles_failed_resume_gracefully`: create a run with no definition snapshot so resume fails, call `spawn_workflow_resume`, join the thread, and assert the run status remains `Failed`.
- **warning** `runkon-flow/src/engine.rs:756` — `restore_completed_step` gained 7 new conditional branches that accumulate `total_cost`, `total_turns`, `total_duration_ms`, `total_input_tokens`, `total_output_tokens`, `total_cache_read_input_tokens`, and `total_cache_creation_input_tokens` from pre-completed steps. No test verifies that these fields are non-zero in the `WorkflowResult` after a resume with populated step metrics. A mismatch in field names or an off-by-one would silently produce wrong totals. Suggestion: Add a unit test that seeds a completed step with non-None metric values, calls `FlowEngine::resume()`, and asserts the returned `WorkflowResult` reflects the accumulated values.
</details>



- [#2609 — CORS configured with overly-permissive allow_methods(Any) and allow_headers(Any)](https://github.com/devinrosen/conductor-ai/issues/2609) — `conductor-web/src/main.rs:770` (warning)

devinrosen (commented):

**Reviewed by:** Security, DRY & Abstraction, Performance, Error Handling, Test Coverage, Architecture

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/engine.rs:1655` — `resume_workflow` constructs `runkon_flow::types::WorkflowExecConfig::default()` for the FlowEngine, silently dropping the caller's shutdown signal. `execute_workflow_standalone` carefully maps all five exec-config fields (lines 1153–1158, including `shutdown`), but `resume_workflow` has no equivalent path because `WorkflowResumeInput` has no `exec_config` or `shutdown` field. Runs launched by `spawn_workflow_resume` and `spawn_heartbeat_resume` are now long-lived FlowEngine executions that cannot be gracefully aborted when the TUI closes or a cancellation is requested externally. Add `shutdown: Option<std::sync::Arc<std::sync::atomic::AtomicBool>>` to both `WorkflowResumeInput` and `WorkflowResumeStandalone`, then thread them into the `WorkflowExecConfig` constructed at line 1655.
</details>
<details>
<summary><b>DRY & Abstraction</b> — 1 issue</summary>

- **warning** `runkon-flow/src/engine.rs:759` — restore_completed_step manually inlines 7 `if let Some(x) { state.total_x += x; }` blocks that duplicate the existing `ExecutionState::accumulate_metrics()` method. The normal execution path (record_step_success, line ~553) already calls `state.accumulate_metrics(...)` for the same fields; the resume path should do the same. Replace lines 759-780 with a single call to `state.accumulate_metrics(step.cost_usd, step.num_turns, step.duration_ms, step.input_tokens, step.output_tokens, step.cache_read_input_tokens, step.cache_creation_input_tokens)`.
</details>
<details>
<summary><b>Error Handling</b> — 2 issues</summary>

- **warning** `conductor-core/src/workflow/engine.rs:1387` — `Err(_)` in `spawn_heartbeat_resume` silently discards the specific `open_database` error. The warning only says 'could not open DB' with no reason — makes it impossible to distinguish a wrong path, permissions failure, or corruption from a log alone. Change `Err(_)` to `Err(db_err)` and include the error in the log: `tracing::warn!(run_id = %run_id, error = %db_err, "spawn_heartbeat_resume: could not open DB to fire stuck-run notification")`.
- **warning** `conductor-tui/src/background.rs:827` — `eprintln!` used for a production warning in `sync_repo` (close_missing_tickets failure) where `tracing::warn!` is the established pattern throughout the codebase. `eprintln!` bypasses structured logging and won't appear in tracing subscribers (file log, OTLP, etc.). Replace `eprintln!("warn: close_missing_tickets failed for {repo_slug}: {e}")` with `tracing::warn!("close_missing_tickets failed for {repo_slug}: {e}")`. Same fix for line 830 (`mark_worktrees_for_closed_tickets`).
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:149` — FlowEngine::resume() is a new public method (~35 lines, branching on resume_ctx and step_map) with no direct unit test in the diff. It is exercised indirectly via the integration test `test_resume_workflow_skips_completed_steps_via_flow_engine`, but edge cases are uncovered: (1) what happens when `state.resume_ctx` is already `Some` on entry, (2) what happens when `persistence.get_steps()` returns an error. Add a unit test covering at least: the error path when persistence.get_steps() fails, and a case where the completed step set is non-empty to verify the skip-set is populated correctly.
- **warning** `conductor-core/src/workflow/engine.rs:71` — spawn_workflow_resume() is a new public function with no test in the diff. The analogous spawn_heartbeat_resume() has test coverage (test_spawn_heartbeat_resume_handles_failed_resume_gracefully), but spawn_workflow_resume() does not. It spawns a detached thread and silently logs on error — the error-logging path is never exercised. Add a test parallel to test_spawn_heartbeat_resume_handles_failed_resume_gracefully that calls spawn_workflow_resume() with a deliberately invalid run_id and asserts the thread completes without panicking.
</details>



- [#2610 — O(n) fan-out idempotency scan in conductor-core InMemoryWorkflowPersistence](https://github.com/devinrosen/conductor-ai/issues/2610) — `conductor-core/src/workflow/persistence_memory.rs:256` (warning)
- [#2611 — `runkon_flow::events::EventSink` leaks into conductor-core's public API](https://github.com/devinrosen/conductor-ai/issues/2611) — `conductor-core/src/workflow/types.rs:338` (warning)
- [#2612 — `WorkflowResumeInput` / `WorkflowResumeStandalone` missing `extra_plugin_dirs` field](https://github.com/devinrosen/conductor-ai/issues/2612) — `conductor-core/src/workflow/engine.rs:1652` (warning)

devinrosen (commented):

**Reviewed by:** Security, Test Coverage, Performance, Architecture, Error Handling, DRY Abstraction

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/engine.rs:0` — `spawn_heartbeat_resume` accepts `workflow_name: String` and `target_label: Option<String>` in its exported signature solely to forward them to `fire_heartbeat_stuck_failed_notification` on the failure path. These are notification-layer concerns leaking into the engine's public API surface. Contrast with `spawn_workflow_resume`, which has no notification params — the asymmetry signals mixed concerns. Callers must thread notification metadata through their stack even though they're invoking an execution primitive. Suggestion: introduce a `SpawnHeartbeatResumeParams` struct to make execution vs. notification parameters explicit, or move the notification call to the single CLI-handler caller that already holds `wf_name` and `label` from the claim tuple.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1330` — `spawn_workflow_resume` and `spawn_heartbeat_resume` both construct `WorkflowResumeStandalone` with the same five constant fields (`model: None`, `from_step: None`, `restart: false`, `shutdown: None`, `workflow_run_id: run_id.clone()`). The only differences are `config`, `db_path`, and `conductor_bin_dir`. If a third spawn variant is added later, this boilerplate will be triplicated. Suggestion: extract a small private helper `fn make_resume_params(config: Config, run_id: String, bin_dir: Option<PathBuf>, db_path: Option<PathBuf>) -> WorkflowResumeStandalone` and call it from both spawn functions.
- **warning** `conductor-tui/src/background.rs:540` — The claim-then-spawn loop for `claim_needs_resume_runs` is copy-pasted verbatim in `conductor-tui/src/background.rs:540-550` and `conductor-cli/src/handlers/workflow.rs:58-68`. Both iterate claimed IDs and call `spawn_workflow_resume(run_id, config.clone(), conductor_bin_dir.clone())`; the only difference is the error log channel (`tracing::warn!` vs `eprintln!`). Suggestion: add a thin `spawn_claimed_runs(claimed: Vec<String>, config: Config, bin_dir: Option<PathBuf>)` helper in the engine module so both call sites reduce to a one-liner and future changes apply once.
</details>
<details>

- **warning** `conductor-core/src/workflow/engine.rs:1324` — `spawn_workflow_resume` is a new `pub fn` with no test in the diff. Its sibling `spawn_heartbeat_resume` (line 1353) has `test_spawn_heartbeat_resume_handles_failed_resume_gracefully` covering the panic-safety and error-logging path. `spawn_workflow_resume` lacks equivalent coverage. Suggestion: add a test analogous to `test_spawn_heartbeat_resume_handles_failed_resume_gracefully`: create a standalone DB, insert a `failed` workflow run with no snapshot, call `spawn_workflow_resume` with a real db_path, join via a timeout or channel, and assert it completes without panicking. Use `make_standalone_db()` from the test helpers to get a real path.
</details>

devinrosen (commented):

**Reviewed by:** DB Migrations, Test Coverage, Error Handling, Architecture, Performance, DRY Abstraction, Security

### Blocking findings

<details>
<summary><b>Architecture</b> — 3 issues</summary>

- **warning** `conductor-core/src/workflow/mod.rs:68` — Execution free functions (`execute_workflow`, `resume_workflow_standalone`, `spawn_claimed_runs`, `spawn_heartbeat_resume`, `validate_resume_preconditions`) are exported at the same level as `WorkflowManager`, creating an asymmetric API surface. State operations require `WorkflowManager::new(&conn).method()`, but execution requires a separate free-function calling convention. No other domain manager (RepoManager, WorktreeManager, AgentManager) exposes this split. This PR adds `resume_workflow_standalone` and `WorkflowResumeStandalone` to the pattern, deepening the inconsistency. Consider adding `WorkflowManager::execute()`, `WorkflowManager::resume_standalone()`, and `WorkflowManager::spawn_claimed_runs()` as methods that delegate to the current free functions internally.
- **warning** `conductor-web/src/routes/workflows.rs:1721` — The web resume handler manually orchestrates `resume_workflow_standalone` followed by a second `open_database` call and direct `notify_workflow()` invocation (lines 1721-1757). The TUI resume path (`modal_dialog.rs:110-141`) calls the same function but fires no notification. Post-execution side effects (notifications) are duplicated in web endpoints but absent in TUI and MCP callers, producing behavioral inconsistency across entry points for the same operation. Notification firing after standalone execution should either be pushed into `resume_workflow_standalone` (accept an optional notification config) or extracted into a shared helper that all callers use.
- **warning** `conductor-tui/src/background.rs:520` — The background watchdog calls `wf_mgr.claim_stuck_workflows()` and `wf_mgr.claim_needs_resume_runs()`, then must separately call `conductor_core::workflow::spawn_claimed_runs()` for each result (lines 519-524 and 537-543). The 'claim then spawn' sequence is a two-step protocol leaked to every caller. Add a `WorkflowManager::claim_and_spawn_stuck_workflows(config, bin_dir)` method that combines `claim_stuck_workflows` + `spawn_claimed_runs` internally. Same for `claim_and_spawn_needs_resume`.
</details>
<details>
<summary><b>DRY Abstraction</b> — 4 issues</summary>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:611` — `claim_heartbeat_stuck_runs` duplicates the WHERE clause of `detect_stuck_workflow_run_ids` (line 371). Both query for `status='running' AND parent_workflow_run_id IS NULL AND NOT EXISTS active steps AND heartbeat threshold`. `claim_stuck_workflows` correctly delegates to `detect_stuck_workflow_run_ids`; `claim_heartbeat_stuck_runs` has its own copy. Extract a private helper returning `Vec<(String, String, Option<String>)>` with the shared WHERE clause.
- **warning** `conductor-core/src/workflow/manager/recovery.rs:1170` — `claim_needs_resume_runs` inlines its own CAS flip (`WHERE id = :id AND status = 'needs_resume'`) instead of calling `cas_flip_run_to_failed`. The only difference from `cas_flip_run_to_failed` is the source status (`needs_resume` vs `running`) and the error string. Generalize `cas_flip_run_to_failed` to `cas_flip_run_to_failed_from(run_id, from_status, error_msg) -> Result<bool>` and have both callers use it.
- **warning** `runkon-flow/src/engine.rs:479` — `record_step_failure` (line 479) and `record_step_skipped` (line 519) each construct a `StepResult` with 8+ fields set to `None`/empty, differing only in `status` and `result_text`. This boilerplate will grow if `StepResult` gains new fields. Add `#[derive(Default)]` to `StepResult` and collapse both to `StepResult { step_name: ..., status: ..., result_text: ..., ..Default::default() }`.
- **warning** `runkon-flow/src/engine.rs:803` — `fetch_child_completion_data` (line 803) and `restore_completed_step` (line 740) both construct a 'completed-from-DB-row' `StepResult` with the identical 10-field pattern: markers, context, child_run_id, structured_output, output_file populated from the DB step row, and `cost_usd: None`, `num_turns: None`, `duration_ms: None`. Extract a `StepResult::from_completed_db_step(step: &WorkflowRunStep, step_key: &str, iteration: u32) -> (StepResult, ContextEntry)` associated function to eliminate the duplicated struct construction and context-push logic.
</details>
<details>
<summary><b>Error Handling</b> — 2 issues</summary>

- **warning** `conductor-core/src/workflow/executors/call_workflow.rs:95` — `.to_str().unwrap_or("")` silently replaces a non-UTF-8 PathBuf with an empty string. When this happens, `load_workflow_by_name()` is called with `working_dir=""` and produces a confusing 'workflow not found' error instead of an 'invalid path encoding' error. The same pattern appears at lines 96, 298, and 299. Return an explicit error: `working_dir.to_str().ok_or_else(|| ConductorError::Workflow(format!("working directory path is not valid UTF-8: {}", working_dir.display())))?`
- **warning** `conductor-web/src/routes/workflows.rs:85` — `.await.unwrap_or(None)` silently discards a `JoinError` from `tokio::task::spawn_blocking`. A panic inside the blocking closure is converted to `None` with no log entry, making the panic invisible to operators. Replace with `.await.unwrap_or_else(|e| { tracing::warn!("wait_for_run_id: spawn_blocking task failed: {e}"); None })` so panics appear in logs.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:465` — `reap_stale_workflow_runs` issues 3 unbatched DB writes per confirmed-dead run (`update_run_failed` + `fail_step_with_message` + `update_workflow_status`), with no enclosing savepoint. The sibling `recover_stuck_steps` immediately above explicitly documents and applies a savepoint for this exact reason: 'Wrap all updates in a savepoint so they commit in one round-trip instead of N separate auto-commit transactions.' With N stale runs this is 3N individual auto-commits. Mirror the `recover_stuck_steps` pattern: open a SAVEPOINT before the for loop, RELEASE on success, ROLLBACK TO + RELEASE on any error.
- **warning** `runkon-flow/src/flow_engine.rs:164` — `FlowEngine::resume()` calls `get_steps()` which fetches every step row for the run (all statuses, all columns) and then immediately discards every non-Completed row. For a large fan-out workflow with 200 steps where 180 are completed, the persistence layer deserializes 200 full rows but only 180 are kept. Add a `get_completed_steps(run_id)` method to the `WorkflowPersistence` trait (or a `get_steps_with_status(run_id, &[status])` variant) so `resume()` can fetch only Completed rows.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:758` — `restore_completed_step` now calls `state.accumulate_metrics(cost_usd, num_turns, duration_ms, ...)` for pre-completed steps during resume, but no test in the diff verifies that non-None metric values are actually accumulated. All resume tests (`resume_skips_completed_steps`, `test_resume_workflow_skips_completed_steps_via_flow_engine`, etc.) seed completed steps with all metric fields as `None`, so the new code path runs but the assertion that `state.total_cost_usd` / token totals are updated is never made. Add a test that seeds a completed step with non-zero `cost_usd` (e.g. `Some(1.23)`) and token fields, calls `engine.resume()`, and asserts that the execution state's accumulated cost/token counters reflect those values after the resume cycle.
</details>



- [#2613 — Domain type `WorkflowRunStep` produces TUI view-model data via `metadata_fields()`](https://github.com/devinrosen/conductor-ai/issues/2613) — `conductor-core/src/workflow/types.rs:0` (warning)
- [#2614 — delete_run_recursive — recursive N+1 query pattern for nested workflow run trees](https://github.com/devinrosen/conductor-ai/issues/2614) — `conductor-core/src/workflow/manager/recovery.rs:999` (warning)
- [#2615 — SENSITIVE_ENV_VARS override is warning-only in script executor](https://github.com/devinrosen/conductor-ai/issues/2615) — `runkon-flow/src/executors/script.rs:209` (warning)

devinrosen (commented):

**Reviewed by:** Error Handling, Performance, Security, DRY & Abstraction, Architecture, Test Coverage

### Blocking findings

<details>

- **warning** `conductor-core/src/workflow/types.rs:604` — `WorkflowResumeInput.event_sinks` uses `runkon_flow::events::EventSink` in a public struct, leaking a runkon-flow type into conductor-core's public API surface. This contradicts the `pub(crate)` isolation on `runkon_bridge.rs` and `rk_types.rs`, which signal that runkon-flow is an implementation detail. Any binary crate that wants to wire a non-empty event sink must now add runkon-flow as a direct dependency. Define a conductor-core-owned `pub trait WorkflowEventSink` and change the field to `Vec<Arc<dyn WorkflowEventSink>>`.
</details>
<details>
<summary><b>DRY & Abstraction</b> — 2 issues</summary>

- **warning** `conductor-core/src/workflow/engine.rs:1094` — ~130-line engine-construction block is duplicated verbatim between `execute_workflow_standalone` (lines ~1094–1225) and `resume_workflow_standalone` (lines ~1604–1730). The resume path even has a comment 'mirrors execute_workflow_standalone'. The duplicate block covers: `rk_def` deserialization, `persistence`/`action_registry`/`item_registry`/`script_env_provider`/`child_runner` construction, the `#[allow(clippy::type_complexity)]` `schema_resolver` closure, and the `FlowEngineBuilder` chain. Extract a helper `fn build_rk_engine_components(...)` that both paths call.
- **warning** `conductor-core/src/workflow/engine.rs:912` — The mutex-poison error string `.map_err(|e| ConductorError::Workflow(format!("db mutex poisoned: {e}")))` is inlined at three separate sites (lines 912, 1233, 1434). Any change to the error message or error type must be made in all three places. Add a one-line helper `fn lock_shared(conn: &Arc<Mutex<Connection>>) -> Result<MutexGuard<Connection>>` and replace all three call sites.
</details>
<details>

- **warning** `conductor-tui/src/background.rs:58` — `.unwrap()` on mutex lock in background polling loop — panics on poison rather than recovering. Lines 58–59 use `lock().unwrap()` for `selected_worktree_id` and `selected_repo_id`, while the rest of the codebase (e.g. `runkon-flow/src/flow_engine.rs:125`) uses `unwrap_or_else(|e| e.into_inner())`. If the TUI main thread panics while holding either guard, the background poll thread crashes silently. Use `selected_worktree_id.lock().unwrap_or_else(|e| e.into_inner()).clone()` (and same for `selected_repo_id`) to match the poison-recovery pattern used elsewhere in the codebase.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:861` — `prev_marker_sets` grows unboundedly across loop iterations. The window check only ever reads the last `stuck_after` entries (`&prev_marker_sets[len - stuck_after..]`), but every iteration's `HashSet<String>` is pushed and retained. A workflow with `max_iterations=500` and `stuck_after=3` will accumulate 500 cloned HashSets by termination. Cap the buffer to `stuck_after` entries using `drain(..prev_marker_sets.len() - cap)` after each push, or use a `VecDeque` with `pop_front`.
- **warning** `conductor-core/src/workflow/persistence_memory.rs:256` — `insert_fan_out_item` checks idempotency with an O(n) linear scan over all fan-out items (`values().find(...)`). For a foreach step with N items, the full dispatch loop is O(N²). The `runkon-flow/src/persistence_memory.rs` sibling already fixes this with a `fan_out_index: HashMap<(String, String), String>` secondary index, making each lookup O(1). Add a matching secondary index to the conductor-core `InMemoryStore`.
</details>
<details>

- **warning** `runkon-flow/src/flow_engine.rs:0` — `FlowEngine::resume()` is a new pub fn with no `#[test]` in the diff. `CountingExecutor`, `make_test_run`, and `make_counting_state` helpers are added to the test module but no `#[test]` function calls `engine.resume()` anywhere in the visible diff. Indirect coverage exists via conductor-core's `test_resume_workflow_skips_completed_steps_via_flow_engine`, but runkon-flow has no in-crate test for the skip-set construction, empty step_map (no-resume_ctx) path, or the guard that rejects a non-None `resume_ctx` on entry. Wire the prepared helpers into at least two tests.
- **warning** `runkon-flow/src/engine.rs:0` — `restore_completed_step` now calls `state.accumulate_metrics()` for each skipped/restored step, but no test in the diff asserts that accumulated metrics (`cost_usd`, `num_turns`, `input_tokens`, etc.) are correctly summed from prior-step records after a partial resume. This is new behavior — before this PR, resumed workflows silently dropped completed-step costs from the run total. Add a test that pre-populates a step with non-zero costs, calls `restore_completed_step`, and asserts the state totals reflect the restored values.
</details>

devinrosen (commented):

**Reviewed by:** Security, Test Coverage, Architecture, Error Handling, DRY/Abstraction, Performance

### Blocking findings

<details>

- **warning** `conductor-cli/src/handlers/workflow.rs:29` — Recovery orchestration sequence duplicated across all three binary crates. This PR replaced the formerly atomic `watchdog_needs_resume_workflows` / `auto_resume_stuck_workflows` calls with separate `claim_*` + `spawn_*` pairs, pushing the assembly of the multi-step sequence (`claim_heartbeat_stuck_runs` → iterate+spawn, `classify_resumable_workflows`, `claim_needs_resume_runs` → `spawn_claimed_runs`) into conductor-cli, conductor-tui/background.rs, and conductor-web/main.rs independently. All three call sites now diverge slightly (e.g. web uses `spawn_workflow_resume` instead of `spawn_claimed_runs`). The orchestration sequence belongs in conductor-core, not in binary crates.
</details>
<details>
<summary><b>DRY/Abstraction</b> — 6 issues</summary>

- **warning** `conductor-core/src/workflow/tests/resumption.rs:142` — `WorkflowResumeInput` struct construction is repeated verbatim in 6 test functions (lines 142-152, 182-192, 231-241, 266-276, 288-298, 344-354). Only `workflow_run_id`, `from_step`, and `restart` vary between calls; the remaining 6 fields are always the same defaults.
- **warning** `conductor-core/src/workflow/tests/resumption.rs:122` — The 12-line `AgentManager::new` → `create_run` → `WorkflowManager::new` → `create_workflow_run` setup sequence is copy-pasted in 5 tests (`test_resume_rejects_completed_run`, `test_resume_rejects_cancelled_run`, `test_resume_rejects_restart_and_from_step_together`, `test_resume_rejects_missing_snapshot`, `test_resume_rejects_nonexistent_from_step`). Only the workflow name and initial status differ.
- **warning** `runkon-flow/src/engine.rs:576` — `ContextEntry` construction is duplicated between `record_step_success` (lines 576–583) and `restore_completed_step` (lines 776–783). Both push the same 6-field struct; the only difference is whether `structured_output`/`output_file` come from a live variable or a `step.field.clone()` call.
- **warning** `runkon-flow/src/engine.rs:804` — `fetch_child_completion_data` iterates `steps` twice with the same `.filter(|s| s.status == WorkflowStepStatus::Completed)` predicate: once with `.iter()` to find `last_completed` (line 804), and once with `.into_iter()` to build `child_steps` (line 819). The completed subset is materialised twice.
- **warning** `runkon-flow/src/persistence_memory.rs:140` — `self.store.lock().map_err(|_| lock_err())?` is repeated on 8+ lines (140, 146, 154, 177, 246, 252, 279, 323, 380, 406, …). The free function `lock_err()` helps name the error but does not reduce the per-call boilerplate.
- **warning** `runkon-flow/src/flow_engine.rs:594` — `AlphaExecutor` and `BetaExecutor` (lines 594–626) are structurally identical: each is a unit struct implementing `ActionExecutor` with a fixed `name()` and a fixed marker string in its `execute()` return. The only difference is a single `&str` literal.
</details>
<details>

- **warning** `conductor-core/src/workflow/runkon_bridge.rs:577` — execute_child() wraps the execute_workflow_standalone error without including the sub-workflow name. When a nested workflow fails, the EngineError surfaced to the parent contains only the inner error string, with no direct indication of which sub-workflow was the culprit.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/helpers.rs:39` — Malformed JSON in `inputs_json` / `blocked_on_json` silently degrades to an empty value rather than propagating an error. While both fields originate from the internal SQLite DB (not an external source), silent data loss can mask database corruption without surfacing to the caller.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/recovery.rs:213` — The rollback path of the new SAVEPOINT in `recover_orphaned_stuck_steps` has no test coverage. The refactor wraps all `update_step_status_full` calls in a savepoint to guarantee atomicity, but only the happy path (all updates succeed → RELEASE) is exercised by existing tests. The error path (ROLLBACK TO SAVEPOINT when one update fails mid-batch) is untested, so a regression that breaks the rollback could go undetected.
</details>



- [#2616 — WorkflowManager directly instantiates AgentManager inside recovery operations — bidirectional manager coupling](https://github.com/devinrosen/conductor-ai/issues/2616) — `conductor-core/src/workflow/manager/recovery.rs:187` (warning)
- [#2617 — prev_marker_sets grows unboundedly — only last stuck_after entries are ever used](https://github.com/devinrosen/conductor-ai/issues/2617) — `runkon-flow/src/engine.rs:861` (warning)

Issue #2635, opened 2026-04-27
Deduplicate CONDUCTOR_OUTPUT parsing between conductor-core and runkon-flow


devinrosen:
## Problem

Both `conductor-core` and `runkon-flow` maintain their own copies of the same agent output protocol:

| Symbol | `runkon-flow` | `conductor-core` |
|---|---|---|
| `CONDUCTOR_OUTPUT_INSTRUCTION` | `src/constants.rs:6` | `src/workflow/constants.rs:40` |
| `ConductorOutput` | `src/helpers.rs:12` | `src/workflow/output.rs:7` |
| `parse_conductor_output()` | `src/helpers.rs:19` | `src/workflow/output.rs:18` |

The implementations are functionally equivalent (find last valid block, strip fences, fix trailing commas, fix escapes, parse JSON). `runkon-flow` already exports these publicly, so `conductor-core` can import them instead of duplicating them.

## Scope

- Add `Default`, `Serialize`, `Deserialize` derives to `runkon-flow::helpers::ConductorOutput` (conductor-core's version has them; downstream may depend on them).
- Update `conductor-core` to import `ConductorOutput`, `parse_conductor_output`, and `CONDUCTOR_OUTPUT_INSTRUCTION` from `runkon-flow` instead of defining its own copies.
- Delete `conductor-core/src/workflow/output.rs` and `conductor-core/src/workflow/constants.rs` (or reduce `constants.rs` to the non-duplicated constants like `STEP_ROLE_FOREACH`, `REGRESSION_*_THRESHOLD`).
- Update `conductor-core/src/workflow/prompt_builder.rs` to use `runkon_flow::constants::CONDUCTOR_OUTPUT_INSTRUCTION`.
- Update `conductor-core/src/workflow/mod.rs` re-exports.

**Out of scope**: `schema_config::extract_output_block`, `strip_trailing_commas`, and `fix_invalid_backslash_escapes` stay in `conductor-core` — they are also used by `parse_structured_output` which is conductor-specific.

## Why now

This should land **before** #2587 (rename `CONDUCTOR_OUTPUT` → `FLOW_OUTPUT`). If we deduplicate first, #2587 becomes a single-crate rename in `runkon-flow` instead of a coordinated two-crate change.

## Migration note

No runtime impact. The protocol strings and parsing logic are identical; this is purely a code cleanup.

PR #2637, opened 2026-04-27
refactor/2635 deduplicate conductor

devinrosen:
- **refactor(#2635): deduplicate ConductorOutput, parse_conductor_output, and CONDUCTOR_OUTPUT_INSTRUCTION**
- **style(#2635): apply cargo fmt formatting fixes**


devinrosen (commented):

**Reviewed by:** Error Handling, Security, Performance, DRY / Abstraction, Architecture, Test Coverage

### Blocking findings (8)

<details>

- **warning** `conductor-core/src/workflow/prompt_builder.rs:5` — `CONDUCTOR_OUTPUT_INSTRUCTION` is imported directly from `runkon_flow::constants` instead of going through the local re-export at `conductor_core::workflow`. The constant is already re-exported by `mod.rs` (`pub use runkon_flow::constants::CONDUCTOR_OUTPUT_INSTRUCTION`), so internal modules should use `use super::CONDUCTOR_OUTPUT_INSTRUCTION`. This bypasses the stable local indirection layer — if the constant moves within `runkon-flow`, two sites need updating instead of one.
</details>
<details>
<summary><b>DRY / Abstraction</b> — 1 issue</summary>

- **warning** `conductor-core/src/workflow/prompt_builder.rs:9` — `substitute_variables_impl` regressed from single-pass to multi-pass: substituted values now get re-scanned and stripped. The old single-pass scan guaranteed that `{{...}}` patterns inside a substituted value (e.g. from `prior_output` containing agent output with JSON object syntax or workflow snippets) were never re-processed. The new implementation runs a second stripping pass over the full result, meaning `{{foo}}` inside a substituted value will be stripped if `foo` is not a known variable — silently corrupting agent prompts. The test `test_substitute_variables_embedded_json_in_value_not_reprocessed` was removed without replacement.
</details>
<details>
<summary><b>Error Handling</b> — 2 issues</summary>

- **warning** `runkon-flow/src/helpers.rs:12` — `.unwrap_or_default()` on `get_steps` silently discards DB errors with no log. The previous implementation explicitly logged `tracing::warn!("build_workflow_summary: failed to fetch steps for run {}: {e}", state.workflow_run_id)`, which made failures diagnosable. The new version will silently produce a 0/0-step summary if the query fails, with no trace in the logs.
- **warning** `conductor-core/src/workflow/prompt_builder.rs:70` — `serde_json::to_string(&state.contexts).unwrap_or_default()` falls back to `""` (empty string — `String::default()`) on serialization failure, not `"[]"`. If this call ever fails, the `{{prior_contexts}}` template variable is injected as an empty string into the agent prompt — silently producing malformed context for downstream steps. No log is emitted. Fix: `.unwrap_or_else(|e| { tracing::warn!("build_variable_map: failed to serialize contexts: {e}"); "[]".to_string() })`
</details>
<details>

- **warning** `conductor-core/src/workflow/prompt_builder.rs:9` — `substitute_variables_impl` replaced a single-pass O(template_len) scanner with an O(N_vars × template_len) multi-pass `str::replace` loop, creating N_vars+1 String allocations instead of one. Agent prompts can be 5–15 KB with ~15 variables, so this is ~15 full-string scans and allocations per prompt build. Additionally, substituted values are now re-scanned — if a substituted value contains `{{key}}` where key is also in vars, it will be doubly substituted, silently corrupting the prompt. The removed test `test_substitute_variables_embedded_json_in_value_not_reprocessed` verified the single-pass guarantee. Fix: restore the single-pass scanner using `String::with_capacity(template.len())` and scanning for `{{ … }}` pairs with `find`.
- **warning** `runkon-flow/src/helpers.rs:116` — `find_max_completed_while_iteration` calls `k.clone()` for every body key on every candidate iteration inside the `all()` check against `skip_set` (a `HashSet<(String, u32)>`). A while loop with 20 iterations and 5 body steps produces 100 String clones in the resume path. Fix: pre-build a `Vec<(String, u32)>` for the current iteration before the `all()` call, or switch to `HashSet<(&str, u32)>` with lifetime-bounded keys.
</details>
<details>

- **warning** `runkon-flow/src/helpers.rs:384` — Trailing-comma test not migrated: `test_parse_conductor_output_trailing_comma_succeeds` was removed from conductor-core but not re-added to runkon-flow. Trailing commas are a documented common LLM artifact and this edge case is now untested. Fix: add a test in the `parse_tests` module that passes `{"markers": ["done"], "context": "trailing comma"}` as input and asserts the parse succeeds.
- **warning** `runkon-flow/src/helpers.rs:384` — Code-fenced block test not migrated: `test_parse_conductor_output_code_fenced` was removed from conductor-core but not re-added to runkon-flow. This covered the case where the JSON payload is wrapped in a ` ```json ` fence inside the CONDUCTOR_OUTPUT block. Fix: add a test that wraps the JSON in a markdown code fence (` ```json\n{...}\n``` `) inside the CONDUCTOR_OUTPUT delimiters and asserts the parse succeeds with correct field values.
</details>



- [#2638 — Duplicate `parse_duration_str` in runkon-flow and conductor-core/workflow_dsl](https://github.com/devinrosen/conductor-ai/issues/2638) — `runkon-flow/src/dsl/parser.rs:1024` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Security, Architecture, Test Coverage, Performance, Error Handling, Dry Abstraction

### Blocking findings (1)

<details>
<summary><b>Dry Abstraction</b> — 1 issue</summary>

- **critical** `runkon-flow/src/helpers.rs:99` — fix_backslash_escapes mishandles valid JSON escape sequences (e.g. \\). It emits a single \ for the first backslash without consuming the second char, causing the next iteration to double the second \. This corrupts valid JSON into unparseable text (\\ becomes \\\\). Since conductor-core now re-exports parse_conductor_output from runkon-flow, the generic output parsing path is affected.
</details>



- [#2639 — Missed constant deduplication](https://github.com/devinrosen/conductor-ai/issues/2639) — `conductor-core/src/workflow/constants.rs:2` (warning)
- [#2640 — Duplicate strip_trailing_commas implementation](https://github.com/devinrosen/conductor-ai/issues/2640) — `conductor-core/src/schema_config/validation.rs:227` (warning)
- [#2641 — Duplicate block extraction logic](https://github.com/devinrosen/conductor-ai/issues/2641) — `conductor-core/src/schema_config/validation.rs:30` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Security, Error Handling, Test Coverage, Performance, Architecture, DRY Abstraction

### Blocking findings (1)

<details>

- **warning** `runkon-flow/src/helpers.rs:57` — parse_conductor_output deserializes into an intermediate serde_json::Value and then manually extracts markers/context, instead of directly deserializing into ConductorOutput. The old conductor-core implementation used direct deserialization (serde_json::from_str::<ConductorOutput>), which avoids the heap-allocated Value tree. Since this diff adds Deserialize to ConductorOutput, direct deserialization is now possible here.
</details>



- [#2642 — Recursive Vec<String> return causes repeated intermediate allocations](https://github.com/devinrosen/conductor-ai/issues/2642) — `runkon-flow/src/helpers.rs:231` (warning)
- [#2643 — Crate boundary violation: conductor-specific concepts in runkon-flow](https://github.com/devinrosen/conductor-ai/issues/2643) — `runkon-flow/src/constants.rs:6` (warning)
- [#2644 — Remaining duplication of string cleanup utilities](https://github.com/devinrosen/conductor-ai/issues/2644) — `conductor-core/src/schema_config/validation.rs:184` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, Security, Performance, Error Handling, Test Coverage, Architecture, DRY/Abstraction

### Blocking findings (13)

<details>
<summary><b>DRY/Abstraction</b> — 9 issues</summary>

- **critical** `conductor-core/src/workflow/executors/call.rs:128` — Agent dispatch setup (load_agent, resolve_schema, load snippets, build_agent_prompt, create_child_run, construct ExecutionContext/ActionParams, registry.dispatch) is duplicated almost line-for-line in parallel.rs. The only meaningful differences are that parallel.rs spawns a thread and omits retry logic. This ~150-line duplication will diverge and cause maintenance bugs when one executor is updated and the other is not.
- **critical** `conductor-core/src/workflow/persistence_sqlite.rs:52` — Every WorkflowPersistence method repeats the exact same three-step ceremony: self.conn.lock().map_err(|_| lock_err())?, WorkflowManager::new(&guard), .map_err(to_engine_err). This pattern appears in 14 of the 15 methods (get_gate_approval is the sole exception). Changing the error mapping strategy requires touching every method.
- **warning** `conductor-core/src/workflow/executors/call.rs:157` — The retry-loop scaffolding (max_attempts = 1 + node.retries, mut last_error, for attempt in 0..max_attempts, insert step as Running, failure update, final handle_on_fail) is structurally identical across call.rs, call_workflow.rs, and script.rs. The loop bodies differ, but the scaffolding is copy-pasted.
- **warning** `conductor-core/src/workflow/engine.rs:1392` — The 7-field metrics accumulation pattern (if let Some(x) = run.field { state.total_... += x }) is inlined in both record_step_success and restore_completed_step, even though accumulate_agent_run already centralizes the same logic at line 177.
- **warning** `conductor-core/src/workflow/engine.rs:1317` — StepResult construction is duplicated 5+ times across the engine (record_step_failure, record_step_skipped, record_step_success, restore_completed_step, bubble_up_child_step_results). The failed and skipped variants are 90% identical; the completed variants also repeat the same 12-field initialization.
- **warning** `conductor-core/src/workflow/manager/steps.rs:166` — The UPDATE workflow_run_steps SET <column> = :val WHERE id = :id pattern is repeated in 10+ setter methods (update_step_child_run_id, set_step_subprocess_pid, set_step_output_file, set_step_gate_info, set_step_parallel_group, set_step_gate_options, and three branches inside update_step_status_full).
- **warning** `conductor-core/src/workflow/manager/queries.rs:197` — Self::STEP_SELECT_WITH_TOKENS.replace("{cols}", &STEP_COLUMNS_WITH_PREFIX) is repeated 5 times across query methods.
- **warning** `conductor-core/src/workflow/manager/queries.rs:394` — The active-worktree guard SQL fragment workflow_runs.worktree_id IS NULL OR worktrees.status = 'active' is hard-coded in 7 separate queries.
- **warning** `conductor-core/src/workflow_dsl/parser.rs:1124` — parse_duration_str is an exact copy of the function in runkon-flow/src/dsl/parser.rs (line 1024). Likewise, ENGINE_INJECTED_KEYS is duplicated between conductor-core/src/workflow/engine.rs and runkon-flow/src/engine.rs. Since conductor-core depends on runkon-flow, or since the DSL is being migrated, this cross-crate duplication should be resolved.
</details>
<details>

- **warning** `runkon-flow/src/helpers.rs:59` — parse_conductor_output JSON parse failure warning lost the 200-char snippet of malformed content that the old implementation included. The raw serde_json error alone is often insufficient to debug LLM output parsing failures.
</details>
<details>

- **warning** `conductor-core/src/workflow/prompt_builder.rs:127` — build_agent_prompt_from_params clones all workflow input values into a HashMap even though template substitution only needs references. For large inputs (file contents, diffs, code snippets) this causes unnecessary heap pressure for every agent prompt.
- **warning** `runkon-flow/src/helpers.rs:267` — find_max_completed_while_iteration performs O(iterations * body_keys) work by calling body_keys.iter().all(...) inside a sequential loop. This scales poorly for while loops with many iterations or large bodies.
- **warning** `conductor-core/src/workflow/prompt_builder.rs:63` — build_prompt_core repeatedly reassigns prompt = format!(...) with the existing prompt as an argument, causing 4-5 full buffer copies of the growing prompt string.
</details>

devinrosen (commented):

**Reviewed by:** Security, Error Handling, Test Coverage, Architecture, DRY Abstraction, Performance

### Blocking findings (10)

<details>

- **warning** `conductor-core/src/workflow/manager/queries.rs:178` — WorkflowManager::resolve_run_context mixes orchestration, filesystem I/O, and cross-domain manager instantiation into a DB query manager. It directly creates WorktreeManager and RepoManager, performs Path::exists() checks, and assembles WorkflowRunContext — none of which belong in a thin persistence wrapper.
</details>
<details>
<summary><b>DRY Abstraction</b> — 4 issues</summary>

- **critical** `conductor-core/src/workflow/prompt_builder.rs:95` — Template-variables display accidentally dropped the `{{`/`}}` braces around placeholder names when refactoring from `format!` to `push_str`. The old output showed `- `{{name}}` = `value`` so agents recognise the placeholder syntax; the new code prints `- `name` = `value``, which is semantically different and less useful to downstream agents.
- **warning** `runkon-flow/src/executors/mod.rs:174` — `record_dispatch_success` doc comment claims it centralises the `persist_completed_step` + `record_step_success` pair duplicated in `call.rs`, `parallel.rs`, and `call_workflow.rs`, but only `call.rs` actually invokes it. `parallel.rs` and `call_workflow.rs` still manually duplicate the pattern.
- **warning** `runkon-flow/src/executors/call_workflow.rs:125` — The resume-success path (~25 lines) and the fresh-success path (~25 lines) are copy-pasted verbatim. They both call `fetch_child_completion_data`, `serialize_or_empty_array`, `persist_completed_step`, `record_step_success`, and iterate `child_steps`. The only differences are the log prefix ("resumed and completed" vs "completed") and the `attempt` value passed to `persist_completed_step`.
- **warning** `runkon-flow/src/executors/call_workflow.rs:150` — Three failure paths (resume-error-resume, resume-error-find, fresh-error-child) each construct a full `StepUpdate { status: Failed, child_run_id: Some(...), result_text: Some(msg.clone()), context_out: None, markers_out: None, retry_count: Some(...), structured_output: None, step_error: Some(msg.clone()) }` literal. `StepUpdate::failed` already exists but hard-codes `child_run_id: None`, so it cannot be reused here.
</details>
<details>
<summary><b>Error Handling</b> — 2 issues</summary>

- **warning** `runkon-flow/src/executors/parallel.rs:223` — `catch_unwind` panic payload is silently discarded by `unwrap_or_else(|_| ...)`. The resulting error message "executor 'X' panicked" contains no root-cause detail, making production panics impossible to debug without additional logging.
- **warning** `conductor-core/src/workflow/manager/queries.rs:984` — `row.get("step_name").unwrap_or_default()` and `row.get("iteration").unwrap_or(0)` silently discard rusqlite errors (e.g., type mismatch or unexpected NULL) without logging, returning empty-string / zero defaults that could propagate incorrect data downstream.
</details>
<details>

- **warning** `runkon-flow/src/executors/mod.rs:174` — record_dispatch_success takes &ActionOutput and clones child_run_id, result_text, structured_output, and markers that were previously moved in call.rs, doubling heap allocations per successful step.
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/steps.rs:29` — set_step_column interpolates the 'column' parameter directly into SQL without parameterization. While all current callers pass hardcoded column names, this is inconsistent with the rest of the codebase and creates latent SQL injection risk if the function is ever called with a user-supplied column name.
</details>
<details>

- **warning** `runkon-flow/src/types.rs:211` — New `impl StepResult` block introduces three public constructors (`failed`, `skipped`, `completed`) without any corresponding unit tests in the diff. The `completed` constructor accepts 10 parameters, making argument-order mistakes easy to introduce and hard to catch without tests.
</details>



- [#2645 — coordinator.rs constructs runkon-flow ExecutionState via raw struct literal](https://github.com/devinrosen/conductor-ai/issues/2645) — `conductor-core/src/workflow/coordinator.rs:0` (warning)
- [#2646 — Prompt builder duplicated across conductor-core and runkon-flow](https://github.com/devinrosen/conductor-ai/issues/2646) — `runkon-flow/src/prompt_builder.rs:0` (warning)
- [#2647 — should_skip allocates on every step lookup](https://github.com/devinrosen/conductor-ai/issues/2647) — `runkon-flow/src/engine.rs:704` (warning)
- [#2648 — Per-step timeout spawns a polling OS thread](https://github.com/devinrosen/conductor-ai/issues/2648) — `runkon-flow/src/executors/call.rs:130` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Test Coverage, Error Handling, Performance, Security, Architecture, DRY Abstraction

### Blocking findings (7)

<details>

- **warning** `conductor-core/src/workflow/manager/steps.rs:36` — WorkflowManager::set_step_column introduces dynamic SQL generation with runtime string validation, deviating from the static SQL manager pattern used consistently by AgentManager, RepoManager, WorktreeManager, ConversationManager, and IssueSourceManager. The DRY benefit (5 call sites) does not justify breaking compile-time safety and pattern consistency.
</details>
<details>

- **critical** `runkon-flow/src/helpers.rs:54` — Duplicate JSON-cleaning helpers: strip_trailing_commas, fix_backslash_escapes, and block-extraction logic are reimplemented here while more robust versions exist in conductor-core/src/schema_config/validation.rs. Since runkon-flow is the lower-level crate, the canonical implementations should live here and conductor-core should delegate to them.
- **warning** `runkon-flow/src/types.rs:233` — StepResult::completed constructor forces callers to pass explicit None for missing metrics fields. Two call sites in engine.rs repeat None, None, None.
</details>
<details>
<summary><b>Performance</b> — 4 issues</summary>

- **critical** `runkon-flow/src/executors/call.rs:207` — Spawns an OS thread per step with a 10 ms polling loop (`std::thread::sleep(Duration::from_millis(10))`). For a 1-hour timeout this causes 360 000 wake-ups per step. This is excessive polling in a tight loop and wastes scheduler time.
- **critical** `runkon-flow/src/engine.rs:432` — `state.persistence.is_run_cancelled()` queries the database on every single workflow node with no throttling or caching. For an N-node workflow this performs N SQLite round-trips before each step.
- **warning** `runkon-flow/src/executors/call_workflow.rs:245` — `child_inputs` is built by cloning `node.inputs` and running variable substitution over every value, but the result is immediately discarded (`let _ = child_inputs;`). `resolve_child_inputs` then repeats the exact same substitution work a second time.
- **warning** `runkon-flow/src/helpers.rs:113` — `find_max_completed_while_iteration` clones every body-key `String` on every iteration check (`skip_set.contains(&(k.clone(), iter))`). For a loop body with M keys and I completed iterations this performs M×I heap allocations.
</details>



- [#2649 — Manager conn visibility inconsistency](https://github.com/devinrosen/conductor-ai/issues/2649) — `conductor-core/src/workflow/manager/mod.rs:22` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Error Handling, DRY/Abstraction, Performance, Test Coverage, Security, Architecture

### Blocking findings (11)

<details>
<summary><b>DRY/Abstraction</b> — 7 issues</summary>

- **critical** `runkon-flow/src/executors/mod.rs:0` — Six shared helper functions were deleted from executors/mod.rs and inlined across call.rs, call_workflow.rs, and parallel.rs, creating heavy duplication: p_err (~30 copies of .map_err(|e| EngineError::Persistence(e.to_string()))), build_execution_context (12-field ExecutionContext literal duplicated), build_inputs_map (HashMap variable-map building duplicated), skip_if_already_completed (should_skip+restore_step+tracing+return pattern duplicated), persist_completed_step (StepUpdate completion persistence duplicated), and begin_retry_attempt (step insertion + StepRetrying event duplicated).
- **critical** `runkon-flow/src/traits/persistence.rs:0` — StepUpdate::completed() and StepUpdate::failed() convenience constructors were removed. The same 7-8 field Completed struct literal is now repeated ~9 times across call.rs, call_workflow.rs, parallel.rs, script.rs, foreach.rs, and gate.rs. The Failed literal is repeated ~12 times.
- **warning** `conductor-core/src/workflow/manager/queries.rs:197` — The STEP_SELECT_EXPANDED LazyLock static was removed. The same Self::STEP_SELECT_WITH_TOKENS.replace("{cols}", &STEP_COLUMNS_WITH_PREFIX) string replacement is now repeated inline at 5 call sites (lines ~197, 256, 279, 308, 767).
- **warning** `conductor-core/src/workflow/manager/steps.rs:0` — The execute_step_sql helper was removed. Every UPDATE wrapper now repeats the exact same self.conn.execute(...)?; Ok(()) boilerplate (~8 occurrences including update_step_child_run_id, set_step_subprocess_pid, set_step_output_file, set_step_gate_info, set_step_parallel_group, set_step_gate_options, reject_gate).
- **warning** `conductor-core/src/workflow/manager/steps.rs:245` — Gate-selection context formatting and serialization logic that previously lived in shared helpers (format_gate_selection_context and serialize_gate_selections) is now inlined directly in approve_gate. The same string-building logic also appears inline in conductor-core/src/workflow/persistence_memory.rs, creating a copy-paste pair that will drift on changes.
- **warning** `runkon-flow/src/types.rs:0` — The Default derive was removed from StepResult. record_step_failure and record_step_skipped in engine.rs now manually initialize every field with the same None/empty defaults, duplicating the initialization pattern.
- **warning** `conductor-core/src/workflow/status.rs:0` — WorkflowStepStatus::is_starting() was removed entirely, and is_terminal() still exists but is no longer used in manager/steps.rs or persistence_memory.rs. Both sites now inline the same explicit match logic (status == Running || status == Waiting and a long matches! for terminal states).
</details>
<details>

- **warning** `runkon-flow/src/helpers.rs:76` — strip_trailing_commas allocates a new String (ws_buf) on the heap for every comma in the input. For JSON text this means repeated small heap allocations in a tight loop.
</details>
<details>
<summary><b>Test Coverage</b> — 3 issues</summary>

- **warning** `conductor-core/src/workflow/manager/queries.rs:981` — fetch_pending_gates now logs warnings via unwrap_or_else when step_name or iteration columns are missing from the result row. This new DB error-handling path is not covered by any test in the diff.
- **warning** `runkon-flow/src/helpers.rs:1504` — parse_conductor_output was switched from manual serde_json::Value extraction to direct derive deserialization. The existing tests validate happy paths but do not exercise edge cases introduced by this stricter parsing (e.g., markers field present but not an array, or context field present but not a string), which now cause a hard parse failure (None) instead of graceful defaulting.
- **warning** `runkon-flow/src/executors/parallel.rs:1464` — New panic payload downcasting logic extracts &str/String messages from thread panics to improve error messages. This behavior change lacks a regression test.
</details>



- [#2650 — SQL limit/offset interpolated directly, defeating prepare_cached](https://github.com/devinrosen/conductor-ai/issues/2650) — `conductor-core/src/workflow/manager/queries.rs:415` (warning)
- [#2651 — Shell command execution via sh -c in notification hooks](https://github.com/devinrosen/conductor-ai/issues/2651) — `conductor-core/src/notification_hooks.rs:197` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Dry Abstraction, Architecture, Error Handling, Security, Test Coverage, Performance

### Blocking findings (22)

<details>
<summary><b>Architecture</b> — 6 issues</summary>

- **critical** `conductor-core/src/workflow/manager/queries.rs:154` — WorkflowManager.resolve_run_context instantiates WorktreeManager and RepoManager inline via their concrete constructors. This tightly couples the workflow domain to worktree/repo domains and violates separation of concerns. Cross-domain orchestration should live in a service/UseCase layer above the managers, or the caller should compose the managers and pass resolved paths.
- **critical** `conductor-core/src/workflow/manager/steps.rs:346` — WorkflowManager.validate_gate_selections mixes domain validation logic (parsing gate options JSON, extracting allowed values, checking selections) with persistence concerns (querying workflow_run_steps.gate_options). A persistence manager should not contain business-rule validation.
- **warning** `conductor-core/src/workflow/manager/queries.rs:154` — WorkflowManager::new(conn) accepts only a Connection, unlike peer managers (WorktreeManager, RepoManager) which store both Connection and Config. This forces resolve_run_context to accept Config ad-hoc and construct dependent managers on the fly, making the API surface inconsistent and harder to test.
- **warning** `runkon-flow/src/engine.rs:904` — Circular module dependency: engine.rs defines ExecutionState and re-exports build_variable_map by delegating to prompt_builder, but prompt_builder.rs imports ExecutionState and ENGINE_INJECTED_KEYS from engine. This reduces cohesion and makes it harder to reason about module boundaries.
- **warning** `runkon-flow/src/executors/call_workflow.rs:275` — The call_workflow executor constructs a placeholder WorkflowDef with empty body, inputs, and always vectors because the engine cannot load workflow definitions, yet the ChildWorkflowRunner trait requires a full WorkflowDef. This is a workaround for an API mismatch.
- **warning** `conductor-core/src/workflow/constants.rs:16` — STEP_ROLE_FOREACH, STEP_ROLE_WORKFLOW, and CONDUCTOR_OUTPUT_INSTRUCTION are duplicated identically in runkon-flow/src/constants.rs. Since conductor-core already depends on runkon-flow, these constants should be re-exported from runkon-flow rather than duplicated, especially on a deduplication branch.
</details>
<details>
<summary><b>Dry Abstraction</b> — 7 issues</summary>

- **critical** `conductor-core/src/workflow/manager/queries.rs:1374` — get_step_failure_heatmap and get_step_retry_analytics share ~80% identical SQL (same FROM/JOIN/WHERE/subquery), differing only in SELECT aggregates and ORDER BY. Fixing a bug in one query (e.g., a status-filter mistake) risks diverging from the other.
- **warning** `conductor-core/src/schema_config/validation.rs:204` — String, Number, and Boolean validation branches repeat the exact same error format with only the type name changing. This is repeated error-mapping boilerplate that could be extracted.
- **warning** `conductor-core/src/schema_config/validation.rs:483` — derive_default_markers hardcodes review-domain field names ('approved', 'findings', 'has_critical_findings') inside generic schema validation code. This couples the generic schema infrastructure to a specific use case and makes the logic invisible to schema authors.
- **warning** `runkon-flow/src/executors/call.rs:151` — The current_execution_id mutex lock/unlock pattern (lock().unwrap_or_else(|e| e.into_inner())) appears twice in execute_call_inner with identical boilerplate.
- **warning** `runkon-flow/src/engine.rs:225` — resolve_child_inputs and apply_workflow_input_defaults both implement required-field checking, boolean defaulting to 'false', and default-value fallback. The logic is duplicated across two functions that operate on the same InputDecl type.
- **warning** `conductor-core/src/workflow/manager/queries.rs:848` — list_all_waiting_gate_steps and list_waiting_gate_steps_for_repo repeat the same column selection, LEFT JOIN agent_runs, JOIN workflow_runs, and base WHERE clause (gate_type IS NOT NULL AND status = 'waiting').
- **warning** `conductor-core/src/schema_config/validation.rs:158` — strip_code_fences in schema_config and the inline fence-stripping logic in runkon-flow/src/helpers.rs parse_conductor_output are two different implementations of the same concept. They behave slightly differently (schema_config handles ```json more robustly).
</details>
<details>

- **warning** `conductor-core/src/workflow/manager/queries.rs:983` — row.get() deserialization errors are silently swallowed with default values (empty string / 0) instead of being propagated. Since the query explicitly selects step_name and iteration, a failure indicates a type mismatch or schema inconsistency that should fail the query rather than return corrupt data downstream.
</details>
<details>
<summary><b>Performance</b> — 5 issues</summary>

- **critical** `conductor-core/src/workflow/manager/queries.rs:647` — get_active_chain_for_run performs up to 5 sequential SQLite queries in a loop (one per depth level), creating an N+1 round-trip pattern for a single logical operation.
- **warning** `conductor-core/src/schema_config/validation.rs:240` — variants.contains(&s.to_string()) allocates a temporary String on every enum validation just to check membership in a Vec<String>.
- **warning** `runkon-flow/src/executors/call.rs:130` — Per-step timeout thread uses a 100 ms polling loop (while start.elapsed() < duration { sleep(100ms); check done }). For typical step timeouts of minutes to hours this generates thousands of unnecessary thread wake-ups.
- **warning** `conductor-core/src/workflow/manager/queries.rs:420` — list_workflow_runs_by_repo_id_filtered interpolates limit and offset directly into the SQL string with format!("... LIMIT {limit} OFFSET {offset}"). When called with varying pagination values this prevents SQLite prepared-statement caching and forces re-parsing/planning on every distinct limit/offset pair.
- **warning** `conductor-core/src/workflow/manager/queries.rs:705` — list_workflow_runs_for_repo interpolates limit directly into the SQL string with format!("... LIMIT {limit}"). Like the paginated variant, this breaks prepared-statement caching for different limit values.
</details>
<details>
<summary><b>Test Coverage</b> — 3 issues</summary>

- **warning** `runkon-flow/src/helpers.rs:95` — `strip_trailing_commas` was promoted from private to `pub` but has no unit tests in the diff. The prior `test_parse_conductor_output_trailing_comma_succeeds` test that exercised trailing comma handling was removed from conductor-core and not migrated to runkon-flow, leaving this edge case uncovered.
- **warning** `runkon-flow/src/traits/persistence.rs:74` — New public constructor `StepUpdate::failed_with_child` is introduced but lacks a unit test in the diff.
- **warning** `conductor-core/src/workflow/manager/queries.rs:984` — Bug fix in `find_stuck_step_names`: replaced silent `unwrap_or_default()` / `unwrap_or(0)` with `unwrap_or_else` that logs warnings when `step_name` or `iteration` columns are missing. No regression test covers this improved error-handling behavior.
</details>



- [#2652 — Library crate contains a binary target](https://github.com/devinrosen/conductor-ai/issues/2652) — `conductor-core/Cargo.toml:0` (warning)

PR #2663, opened 2026-04-27
refactor(#2579): introduce StepSuccess struct to replace 17 positional args

devinrosen:
record_step_success previously took 17 positional arguments, making call sites verbose and error-prone. Introduce a StepSuccess struct that groups all step-success metadata and update record_step_success to accept &StepSuccess.

Also update StepResult::completed to take &StepSuccess so the struct is used consistently throughout the success-path pipeline.

Files changed:
- types.rs: add StepSuccess struct; refactor StepResult::completed
- engine.rs: refactor record_step_success signature
- executors/mod.rs: update record_dispatch_success
- executors/call_workflow.rs: build StepSuccess at call site
- executors/foreach.rs: simplify record_foreach_step_success wrapper
- executors/script.rs: build StepSuccess at both call sites

### DIFF
--- a/.conductor/agents/review-aggregator.md
+++ b/.conductor/agents/review-aggregator.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-haiku-4-5
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a review aggregator. Your job is to aggregate findings from multiple parallel code reviewers, determine whether the PR is ready to merge, and produce a structured output. A subsequent script step will build the review body and submit the formal GitHub PR review.
--- a/.conductor/agents/review-architecture.md
+++ b/.conductor/agents/review-architecture.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-sonnet-4-6
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a senior software architect reviewing a pull request on a Rust project.
--- a/.conductor/agents/review-db-migrations.md
+++ b/.conductor/agents/review-db-migrations.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-haiku-4-5
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a database migration reviewer working on a Rust project using SQLite with WAL mode.
--- a/.conductor/agents/review-dry-abstraction.md
+++ b/.conductor/agents/review-dry-abstraction.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-sonnet-4-6
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a code quality reviewer focused on DRY principles and abstraction in a Rust codebase.
--- a/.conductor/agents/review-error-handling.md
+++ b/.conductor/agents/review-error-handling.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-haiku-4-5
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are an error-handling reviewer working on a Rust project.
--- a/.conductor/agents/review-performance.md
+++ b/.conductor/agents/review-performance.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-sonnet-4-6
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a performance-focused code reviewer working on a Rust project.
--- a/.conductor/agents/review-security.md
+++ b/.conductor/agents/review-security.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-sonnet-4-6
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a security-focused code reviewer working on a Rust CLI/TUI tool that manages git repos and spawns AI agents.
--- a/.conductor/agents/review-test-coverage.md
+++ b/.conductor/agents/review-test-coverage.md
@@ -1,6 +1,7 @@
 ---
 role: reviewer
-model: claude-haiku-4-5
+runtime: kimi
+model: moonshotai/Kimi-K2.6
 ---
 
 You are a test coverage reviewer working on a Rust project.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2123,6 +2123,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
+name = "hermit-abi"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
+
+[[package]]
 name = "hex"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3309,6 +3315,16 @@ checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
 dependencies = [
  "autocfg",
  "libm",
+]
+
+[[package]]
+name = "num_cpus"
+version = "1.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91df4bbde75afed763b708b7eee1e8e7651e02d97f6d5dd763e89367e957b23b"
+dependencies = [
+ "hermit-abi",
+ "libc",
 ]
 
 [[package]]
@@ -4525,8 +4541,10 @@ dependencies = [
  "serde_json",
  "tempfile",
  "thiserror 2.0.18",
+ "threadpool",
  "tracing",
  "ulid",
+ "wait-timeout",
 ]
 
 [[package]]
@@ -5825,6 +5843,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f60246a4944f24f6e018aa17cdeffb7818b76356965d03b07d6a9886e8962185"
 dependencies = [
  "cfg-if",
+]
+
+[[package]]
+name = "threadpool"
+version = "1.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d050e60b33d41c19108b32cea32164033a9013fe3b46cbd4457559bfbf77afaa"
+dependencies = [
+ "num_cpus",
 ]
 
 [[package]]
--- a/conductor-core/src/workflow/runkon_bridge.rs
+++ b/conductor-core/src/workflow/runkon_bridge.rs
@@ -493,22 +493,21 @@ impl ConductorChildWorkflowRunner {
 impl runkon_flow::engine::ChildWorkflowRunner for ConductorChildWorkflowRunner {
     fn execute_child(
         &self,
-        child_def: &runkon_flow::dsl::WorkflowDef,
+        workflow_name: &str,
         parent_state: &runkon_flow::engine::ExecutionState,
         params: runkon_flow::engine::ChildWorkflowInput,
     ) -> runkon_flow::engine_error::Result<runkon_flow::types::WorkflowResult> {
-        // Load the real workflow definition from disk. The caller passes a placeholder
-        // WorkflowDef with body=[] — the child runner is responsible for resolving the
+        // Load the real workflow definition from disk. The runner resolves the
         // actual definition by name from the worktree/repo .conductor/workflows/ directory.
         let core_def = runkon_flow::dsl::load_workflow_by_name(
             &parent_state.worktree_ctx.working_dir,
             &parent_state.worktree_ctx.repo_path,
-            &child_def.name,
+            workflow_name,
         )
         .map_err(|e| {
             EngineError::Workflow(format!(
                 "failed to load sub-workflow '{}': {e}",
-                child_def.name
+                workflow_name
             ))
         })?;
 
@@ -554,7 +553,7 @@ impl runkon_flow::engine::ChildWorkflowRunner for ConductorChildWorkflowRunner {
                 ConductorError::WorkflowCancelled => EngineError::from(e),
                 other => EngineError::Workflow(format!(
                     "child workflow '{}' failed: {other}",
-                    child_def.name
+                    workflow_name
                 )),
             })?;
 
--- a/runkon-flow/Cargo.toml
+++ b/runkon-flow/Cargo.toml
@@ -18,6 +18,8 @@ chrono = { version = "0.4", features = ["serde"] }
 tracing = "0.1"
 dirs = "6"
 tempfile = "3"
+threadpool = "1"
+wait-timeout = "0.2"
 
 [[test]]
 name = "basic_flow"
--- a/runkon-flow/src/engine.rs
+++ b/runkon-flow/src/engine.rs
@@ -170,6 +170,8 @@ impl ExecutionState {
     }
 
     /// Accumulate individual metrics into this execution state.
+    ///
+    /// Returns `true` if at least one metric was present and added.
     #[allow(clippy::too_many_arguments)]
     pub fn accumulate_metrics(
         &mut self,
@@ -180,28 +182,37 @@ impl ExecutionState {
         output_tokens: Option<i64>,
         cache_read: Option<i64>,
         cache_create: Option<i64>,
-    ) {
+    ) -> bool {
+        let mut changed = false;
         if let Some(c) = cost {
             self.total_cost += c;
+            changed = true;
         }
         if let Some(t) = turns {
             self.total_turns += t;
+            changed = true;
         }
         if let Some(d) = duration {
             self.total_duration_ms += d;
+            changed = true;
         }
         if let Some(t) = input_tokens {
             self.total_input_tokens += t;
+            changed = true;
         }
         if let Some(t) = output_tokens {
             self.total_output_tokens += t;
+            changed = true;
         }
         if let Some(t) = cache_read {
             self.total_cache_read_input_tokens += t;
+            changed = true;
         }
         if let Some(t) = cache_create {
             self.total_cache_creation_input_tokens += t;
+            changed = true;
         }
+        changed
     }
 
     /// Persist the current accumulated metrics to the workflow run row.
@@ -563,7 +574,7 @@ pub fn record_step_success(
     step_key: String,
     success: crate::types::StepSuccess,
 ) {
-    state.accumulate_metrics(
+    let metrics_changed = state.accumulate_metrics(
         success.cost_usd,
         success.num_turns,
         success.duration_ms,
@@ -573,9 +584,11 @@ pub fn record_step_success(
         success.cache_creation_input_tokens,
     );
 
-    // Best-effort mid-run metrics flush — non-fatal
-    if let Err(e) = state.flush_metrics() {
-        tracing::warn!("Failed to flush mid-run metrics: {e}");
+    // Best-effort mid-run metrics flush — non-fatal, only if something changed
+    if metrics_changed {
+        if let Err(e) = state.flush_metrics() {
+            tracing::warn!("Failed to flush mid-run metrics: {e}");
+        }
     }
 
     let step_result = StepResult::completed(&success);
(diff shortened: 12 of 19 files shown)


devinrosen (commented):

**Reviewed by:** Test Coverage, DRY Abstraction, Performance, Error Handling, Security, Architecture

### Blocking findings (6)

<details>

- **warning** `runkon-flow/src/types.rs:259` — `StepResult::completed_without_metrics` is the only public `StepResult` constructor that still takes positional arguments. It internally creates a temporary `StepSuccess` to delegate to `completed(&StepSuccess)`, making the `StepResult` API surface inconsistent: one constructor uses the new parameter-object pattern, the other does not.
</details>
<details>
<summary><b>DRY Abstraction</b> — 1 issue</summary>

- **warning** `runkon-flow/src/engine.rs:530` — Three intermediate clone variables (`markers_for_ctx`, `structured_output_for_ctx`, `output_file_for_ctx`) exist only because `push_context_entry` takes owned fields instead of `&StepSuccess`. After this refactor, `push_context_entry` is the last call site that forces this field-by-field splitting.
</details>
<details>

- **warning** `runkon-flow/src/executors/call_workflow.rs:299` — Log message omits child run ID when sub-workflow completes but all_succeeded=false. The run ID is persisted in the DB via `failed_with_child`, but is absent from the tracing::warn! call, making it hard to locate the child failure from logs alone.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:530` — Double-clone of `markers`, `structured_output`, `output_file`, and `context` on every step completion. `record_step_success` takes `&StepSuccess` by reference, so lines 530–532 and 540 clone those fields for `push_context_entry`, while `StepResult::completed` (types.rs:291–295) clones the same fields again from the same reference. For callers like `record_dispatch_success` that already clone these strings from `output` to construct `StepSuccess`, the total clone count reaches three per field. `context` and `structured_output` can be multi-KB strings (full agent output, JSON blobs), making this a real per-step allocation cost.
</details>
<details>
<summary><b>Security</b> — 2 issues</summary>

- **warning** `runkon-flow/src/executors/script.rs:200` — SENSITIVE_ENV_VARS check logs a warning but does not block the insert: `LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, and `PATH` are still written into `env_vars` at line 208. A workflow author can override these to affect the subprocess (e.g. inject a shared library via LD_PRELOAD). The warning creates a misleading appearance of protection that does not exist.
- **warning** `runkon-flow/src/executors/script.rs:182` — `DYLD_LIBRARY_PATH` is absent from the SENSITIVE_ENV_VARS list. On macOS, `DYLD_LIBRARY_PATH` can be used to inject shared libraries into a subprocess the same way `LD_LIBRARY_PATH` can on Linux, and it is commonly used alongside `DYLD_INSERT_LIBRARIES`. The omission is inconsistent with the intent of the list.
</details>



- [#2664 — `completed_without_metrics` is a positional-arg holdover that contradicts the StepSuccess refactor](https://github.com/devinrosen/conductor-ai/issues/2664) — `runkon-flow/src/types.rs:259` (warning)

devinrosen (commented):

**Reviewed by:** Error Handling, Test Coverage, Security, Architecture, DRY Abstraction, Performance

### Blocking findings (4)

<details>

- **warning** `runkon-flow/src/types.rs:276` — `StepResult::completed` (and `completed_without_metrics`) take a standalone `step_name: &str` parameter that is always identical to `success.step_name`. The `StepSuccess` struct was introduced to be the single source of truth for step completion data, but the factory methods still require callers to extract and re-pass `step_name` externally — defeating that goal and creating a sync hazard. Suggestion: remove the `step_name: &str` parameter from both methods and use `success.step_name.clone()` directly.
</details>
<details>

- **warning** `runkon-flow/src/types.rs:255` — `completed_without_metrics` no longer delegates to `completed`, duplicating 9 identical struct field assignments. The two methods are identical except for setting cost_usd/num_turns/duration_ms to None. Suggestion: have `completed_without_metrics` call `Self::completed(success)` and then zero out the three metric fields, restoring the delegation pattern that existed before the refactor.
- **warning** `runkon-flow/src/types.rs:271` — Both `completed` and `completed_without_metrics` take an explicit `step_name: &str` parameter that duplicates `success.step_name` — every call site passes `success.step_name` (or its equivalent) as the first argument. The parameter is redundant and forces callers to keep two values in sync. Suggestion: remove the separate `step_name` parameter and read it from `success.step_name` directly; update the call site from `StepResult::completed(&success.step_name, success)` to `StepResult::completed(success)`.
</details>
<details>

- **warning** `runkon-flow/src/executors/script.rs:200` — Security behavior fix (sensitive env var filtering now uses `continue` to actually skip vars) has no regression test. Before this diff, sensitive vars like PATH/DYLD_LIBRARY_PATH were warned about but still inserted into `env_vars`; now they are skipped. DYLD_LIBRARY_PATH was also silently added to the sensitive list. Neither behavior is covered by a test in this diff. Suggestion: add a unit test asserting sensitive vars are absent from the resolved env map, or extract the filtering loop into a pure helper and test it directly.
</details>

devinrosen (commented):

**Reviewed by:** Security, Error Handling, Architecture, Test Coverage, Performance, DRY Abstraction

### Blocking findings (5)

<details>

- **warning** `runkon-flow/src/executors/call_workflow.rs:242` — `ChildWorkflowRunner::execute_child` takes `child_def: &WorkflowDef` but both call sites (call_workflow.rs:242 and foreach.rs:373) pass a stub value with only `name` set and an empty body. The inline comment explicitly states 'We pass an empty WorkflowDef as a placeholder and let the runner resolve it by name.' This is a misleading API contract: callers must knowingly violate the parameter's semantic promise, and implementations must know to ignore it. The trait should take `workflow_name: &str` instead of `&WorkflowDef`, making the contract honest.
</details>
<details>

- **warning** `runkon-flow/src/types.rs:200` — `StepSuccess.step_key` and `.iteration` are never read by `StepResult::completed()` or `completed_without_metrics()`, yet the struct is used as input to both functions. `fetch_child_completion_data` (engine.rs:777-786) must construct a `StepSuccess` with `step_key` defaulting to empty string and `iteration` to 0 solely to call `completed_without_metrics()`. The struct docstring says it is "passed to `record_step_success`", but it has leaked into `StepResult`'s API, conflating two concerns: state-map insertion key (`step_key`, `iteration`) with step output data (everything else).
- **warning** `runkon-flow/src/engine.rs:722` — `restore_completed_step` sets `success.step_key` in the `StepSuccess` struct (line 722), but the map insertion at line 734 uses the local `step_key: &str` variable directly rather than `success.step_key`. This creates an inconsistency with `record_step_success` (line 531), which routes the insertion through `success.step_key`. A future reader familiar with `record_step_success` would expect `success.step_key` to control insertion here too, but it is structurally dead — set but never consumed through the struct.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:514` — record_step_success takes &StepSuccess, causing context, markers, structured_output, and output_file to be cloned twice per successful step: once inside StepResult::completed(&success) (line 530) and again inside push_context_entry(state, success) (line 533). The context and structured_output fields can be 10–100 KB for agent steps, making the double clone measurable in high-throughput workflows.
</details>
<details>

- **warning** `runkon-flow/src/types.rs:255` — StepResult::completed_without_metrics is a public function whose implementation changed: it now calls completed(success) (which copies metric fields) and then zeroes them, rather than passing None directly. No test in the diff verifies that a StepSuccess with non-None metrics produces a StepResult with cost_usd/num_turns/duration_ms all None. The 'metric fields are ignored' contract is untested.
</details>



- [#2665 — TLS trust-anchor env vars not blocked in SENSITIVE_ENV_VARS list](https://github.com/devinrosen/conductor-ai/issues/2665) — `runkon-flow/src/executors/script.rs:182` (warning)
- [#2666 — `accumulate_metrics` retains too-many-arguments suppression](https://github.com/devinrosen/conductor-ai/issues/2666) — `runkon-flow/src/engine.rs:126` (warning)
- [#2667 — `persist_completed_step` retains too-many-arguments suppression](https://github.com/devinrosen/conductor-ai/issues/2667) — `runkon-flow/src/executors/mod.rs:103` (warning)

devinrosen (commented):

**Reviewed by:** Test Coverage, Error Handling, Architecture, Security, DRY Abstraction, Performance

### Blocking findings (4)

<details>

- **warning** `runkon-flow/src/types.rs:198` — StepSuccess docstring claims `iteration` is excluded from the struct, but `iteration: u32` is a field at line 214 and is actively read by `record_step_success` to populate `ContextEntry.iteration`. The comment was likely written to describe design intent (keeping the step_key tuple separate) but was not updated when `iteration` was added. Future maintainers will be misled about the type's contract. Suggested fix: change the docstring to "Does not include `step_key` — that is an execution bookkeeping concern kept as a separate parameter. `iteration` is included because it is needed to populate `ContextEntry`."
</details>
<details>
<summary><b>DRY Abstraction</b> — 1 issue</summary>

- **warning** `runkon-flow/src/engine.rs:553` — Identical `ContextEntry { ... }` struct literal appears at `record_step_success` (line 553) and `restore_completed_step` (line 751). The deleted `push_context_entry` helper was the previous dedup point; its removal left the same 6-field construction inlined twice. Suggested fix: implement `From<StepSuccess> for ContextEntry` in `types.rs` (mapping the 6 shared fields). Both call sites become `state.contexts.push(success.into())`, and any future `ContextEntry` field additions from `StepSuccess` only need updating in one place.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:741` — Unnecessary clone of `context` before moving into `StepSuccess`. `context` is not used after line 741, so `context.clone()` allocates a full copy of what can be a large LLM output string. `StepResult::completed_without_metrics(&success)` then clones it a second time, and `success.context` is finally moved to `ContextEntry`. One of these copies is avoidable. Suggested fix: change `context: context.clone()` to `context: context` on line 741 (move instead of clone). With 50 completed steps, that is 50 unnecessary copies of potentially large context strings.
- **warning** `runkon-flow/src/engine.rs:799` — Double-cloning in `fetch_child_completion_data`: each DB step row clones step_name, result_text, child_run_id, structured_output, and output_file into an intermediate `StepSuccess`, then `StepResult::completed_without_metrics(&success)` clones all of them again because it takes `&StepSuccess`. Since no `ContextEntry` is needed in this function, the `StepSuccess` intermediary serves no purpose beyond the existing call to `completed_without_metrics`. Suggested fix: build `StepResult` directly in the `map` closure instead of routing through an intermediate `StepSuccess`, or add `StepResult::from_success(success: StepSuccess)` that consumes by value.
</details>

devinrosen (commented):

**Reviewed by:** DB Migrations, Security, Test Coverage, DRY/Abstraction, Architecture, Performance, Error Handling

### Blocking findings (4)

<details>
<summary><b>DRY/Abstraction</b> — 2 issues</summary>

- **warning** `runkon-flow/src/engine.rs:534` — `record_step_success` manually constructs `StepResult` (lines 536-551) in a body that is field-for-field identical to `StepResult::completed(&success)`, which was added precisely for this call site. The comment justifying the manual construction ('avoid double-cloning') is incorrect: both approaches clone each field exactly once from `&StepSuccess`, then move `success` into `ContextEntry` via `.into()`. Because the factory method is bypassed here, adding any field to `StepSuccess`/`StepResult` requires two updates instead of one, defeating the purpose of the factory.
- **warning** `runkon-flow/src/executors/parallel.rs:290` — `ContextEntry` is constructed manually from raw `ActionOutput` fields and local variables instead of via the `From<StepSuccess>` trait defined in `types.rs:301-312`. The parallel executor intentionally skips `record_step_success` (sub-agent results should not appear as top-level step results), but the `ContextEntry` push still repeats the exact field mapping that `From<StepSuccess>` encodes. `output_file: None` is hardcoded rather than taken from `output`, which will silently diverge if parallel sub-agents ever start producing output files.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:541` — Unnecessary clone of `result_text` (and `child_run_id` at line 547) in `record_step_success`. Both fields are cloned into `StepResult` but neither is present in `ContextEntry` (see `From<StepSuccess> for ContextEntry` in types.rs:298-308), so the originals inside `success` are dropped unconsumed after `success.into()`. Before this PR both fields were moved directly into `StepResult` without cloning. `result_text` can hold the full LLM agent output (potentially tens of KB), making this a meaningful per-step allocation regression for long-running workflows.
</details>
<details>

- **warning** `conductor-core/src/workflow/helpers.rs:26` — `format_gate_selection_context` is a new `pub fn` with a loop (builds a bulleted-list string) but has no unit test in the diff. The helpers.rs test block only covers `parse_gate_options`.
</details>



- [#2668 — persist_metrics() is the same positional-arg problem this PR solved — needs a struct](https://github.com/devinrosen/conductor-ai/issues/2668) — `conductor-core/src/workflow/persistence_sqlite.rs:226` (warning)
- [#2669 — GateParams embeds runkon_flow::dsl::ApprovalMode directly, bypassing the bridge layer](https://github.com/devinrosen/conductor-ai/issues/2669) — `conductor-core/src/workflow/executors/gate_resolver.rs:10` (warning)
- [#2670 — foreach.rs: `still_waiting` Vec reallocated on every item completion](https://github.com/devinrosen/conductor-ai/issues/2670) — `runkon-flow/src/executors/foreach.rs:516` (warning)
- [#2671 — recovery.rs: per-step `mark_step_terminal()` calls inside savepoint](https://github.com/devinrosen/conductor-ai/issues/2671) — `conductor-core/src/workflow/manager/recovery.rs:250` (warning)

devinrosen (commented):

**Reviewed by:** Test Coverage, DRY Abstraction, Error Handling, Performance, Architecture, Security

### Blocking findings (4)

<details>

- **warning** `runkon-flow/src/executors/parallel.rs:290` — Successful parallel branches manually replicate the body of `record_step_success` (accumulate_metrics + flush_metrics + contexts.push) inline instead of calling it. Every other executor routes through `record_step_success`; this path is the only outlier. Any future addition to `record_step_success` (e.g. a StepCompleted event) will silently be absent from parallel steps. Suggestion: call `record_step_success(state, agent_step_key, StepSuccess { ... })` for each successful branch, then build the synthetic merged-marker `StepResult` with the `parallel:<group_id>` key afterwards.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:769` — `fetch_child_completion_data` constructs a `StepResult` via a direct struct literal instead of `StepResult::completed_without_metrics()`. The doc-comment on that constructor explicitly calls out the 'bubble-up from a child workflow' case — which is exactly what this function is. The sibling call site `restore_completed_step` uses the new API correctly; this call site does not. Suggestion: build a `StepSuccess` from the `s` row fields (with `..StepSuccess::default()` for metrics) and pass it to `StepResult::completed_without_metrics(&success)`.
- **warning** `runkon-flow/src/executors/parallel.rs:290` — `StepSuccess` (15 fields) is constructed purely as a transient intermediary to call `.into::<ContextEntry>()`, which discards 9 of those fields (result_text, all 7 token/cost/turn metrics, child_run_id). Metrics are accumulated separately on the lines immediately following, so populating them on `StepSuccess` here is misleading and wasted effort. Suggestion: replace with a direct `ContextEntry { step: pr.agent_name.clone(), iteration, context: context.clone(), markers: output.markers.clone(), structured_output: output.structured_output.clone(), output_file: None }` construction.
</details>
<details>

- **warning** `runkon-flow/src/types.rs:743` — `impl From<StepSuccess> for ContextEntry` has no direct unit test. The impl maps 6 fields including a rename (`step_name` → `step`); only the `context` field is indirectly exercised via `sensitive_env_vars_are_blocked`. A field-mapping error (swap, omission, or wrong rename) would not be caught. Suggestion: add a unit test constructing a `StepSuccess` with distinct non-default values for all 6 ContextEntry fields, calling `.into()`, and asserting each field — particularly verifying `step == step_name`.
</details>



- [#2672 — accumulate_metrics still has 7 positional args after StepSuccess was introduced](https://github.com/devinrosen/conductor-ai/issues/2672) — `runkon-flow/src/engine.rs:126` (warning)

devinrosen (commented):

**Reviewed by:** Test Coverage, Dry Abstraction, Security, Performance, Architecture, Error Handling

### Blocking findings (3)

<details>

- **warning** `runkon-flow/src/executors/foreach.rs:71` — `ForeachParentCtx::make_child_state()` constructs a full `ExecutionState` struct literal naming every field. This disperses child-state construction logic into the foreach executor rather than keeping it in the engine module where `ExecutionState` is defined. Fields added to `ExecutionState` after the initial implementation (e.g. `schema_resolver`, `triggered_by_hook`, `block_output`, `block_with`, `current_execution_id`) appear as bare `None`/`false`/`vec![]` assignments, confirming the maintenance burden this creates. Suggestion: Add a `fork_child(&self, cancellation: CancellationToken) -> ExecutionState` factory method to `ExecutionState` in `engine.rs`; `ForeachParentCtx::make_child_state` becomes a one-liner.
</details>
<details>
<summary><b>Dry Abstraction</b> — 1 issue</summary>

- **warning** `runkon-flow/src/executors/parallel.rs:302` — ActionOutput → StepSuccess field mapping is duplicated verbatim here and in executors/mod.rs:185. Both constructions clone the same 11 fields from ActionOutput; the only differences are the step_name source and whether output_file is a param or None. Suggestion: Add a constructor to StepSuccess — e.g. `StepSuccess::from_action_output(output: &ActionOutput, step_name: String, context: String, iteration: u32, output_file: Option<String>) -> Self` — and use it in both mod.rs:record_dispatch_success and the parallel success arm. This keeps the field mapping in one place and makes it easier to add new ActionOutput fields in the future.
</details>
<details>

- **warning** `runkon-flow/src/executors/parallel.rs:297` — `output.markers` is cloned twice in the success branch: once via `.iter().cloned()` into `merged_markers` (line 297) and again via `.clone()` into `StepSuccess` (line 312). The same double-clone applies to `result_text`, `child_run_id`, and `structured_output`. Root cause: the loop iterates by reference (`for pr in &results`), which prevents moving owned values. `results` is only used as `.len()` after the loop, so consuming it is a straight swap. Fix: capture `let results_count = results.len();` before the loop, then change `for pr in &results { match &pr.result {` to `for pr in results { match pr.result {` and use `results_count` in place of `results.len()` on the total_agents line.
</details>



- [#2673 — NODE_OPTIONS and other interpreter injection vars absent from SENSITIVE_ENV_VARS blocklist](https://github.com/devinrosen/conductor-ai/issues/2673) — `runkon-flow/src/executors/script.rs:182` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Architecture, Test Coverage, Error Handling, Security, DRY Abstraction, Performance

### Blocking findings (7)

<details>

- **warning** `runkon-flow/src/executors/parallel.rs:238` — `let _ = tx.send(...)` silently discards channel send errors. If the receiver is dropped (e.g., main thread panic), the failure is invisible and the branch result is lost without any log.
</details>
<details>
<summary><b>Performance</b> — 5 issues</summary>

- **critical** `runkon-flow/src/executors/foreach.rs:514` — Spawns a new OS thread for every foreach work item via std::thread::spawn. Even with max_parallel bounding concurrency, processing N items creates N threads over time, causing severe syscall overhead and OS scheduler pressure.
- **critical** `runkon-flow/src/executors/script.rs:257` — cmd.status() blocks the workflow engine thread indefinitely and ignores ScriptNode.timeout. A hung or long-running script will stall the entire workflow run with no mechanism for cancellation or timeout.
- **warning** `runkon-flow/src/executors/foreach.rs:451` — After every item completion, the dispatch loop drains the entire waiting Vec and re-scans each entry with is_eligible, which checks all dependencies against terminal_ids. In ordered workflows with long dependency chains this becomes O(N×M) per completion and O(N²) overall.
- **warning** `runkon-flow/src/executors/script.rs:215` — Creates NamedTempFile for stdout and stderr, then reads them back from disk. This adds unnecessary filesystem I/O (create, write, read, delete) when std::process::Command::output() captures both streams in memory without leaking to the terminal.
- **warning** `runkon-flow/src/engine.rs:577` — record_step_success unconditionally calls flush_metrics(), which performs a DB write on every completed step. For workflows with many steps this updates the same workflow run row N times in addition to the final flush at run completion.
</details>
<details>

- **warning** `runkon-flow/src/engine.rs:128` — New public method `ExecutionState::fork_child` has no unit test in the diff. It contains non-trivial logic for cloning shared state (persistence, registries, workflow identity) and resetting runtime accumulators (inputs, step_results, contexts, position, totals) so the child starts with a clean slate. Without coverage, a future refactor could accidentally break child isolation (e.g., copying inputs instead of clearing them).
</details>

devinrosen (commented):

**Reviewed by:** DRY Abstraction, Security, Test Coverage, Architecture, Performance, Error Handling

### Blocking findings (8)

<details>
<summary><b>DRY Abstraction</b> — 1 issue</summary>

- **warning** `runkon-flow/src/executors/script.rs:0` — Four nearly identical script-step failure blocks repeat the same 17-line sequence: format error message, tracing::warn!, persistence.update_step with a verbose StepUpdate { status: Failed, ... } struct literal, and apply_script_on_fail. The StepUpdate::failed(err_msg, 0) constructor already exists and produces exactly this shape, yet it is not used here. A single helper would collapse ~80 lines to ~4 lines and prevent drift if failure bookkeeping changes.
</details>
<details>

- **warning** `runkon-flow/src/executors/script.rs:309` — pipe.read_to_end() and String::from_utf8() errors are silently discarded via .ok()? / .ok() without any log, making it impossible to debug missing script stdout/stderr in production. The previous temp-file implementation logged these failures with tracing::warn!.
</details>
<details>

- **critical** `runkon-flow/src/executors/script.rs:197` — Pipe deadlock risk: both stdout and stderr are piped (Stdio::piped()) but are only read AFTER child.wait()/wait_timeout() returns. If the child writes more than the OS pipe buffer size (~64KB on Linux) to either stream, it will block forever. The parent is simultaneously blocked in wait(), creating a deadlock. This is most severe when node.timeout is None (the default), causing an unbounded hang.
- **warning** `runkon-flow/src/executors/script.rs:318` — Unnecessary heap allocation on successful script execution: stderr is fully read into a String (lines 318-326) but is never referenced in the success path. The old code deleted the stderr temp file unread on success, avoiding this I/O and allocation entirely.
- **warning** `runkon-flow/src/executors/foreach.rs:37` — Wasteful clone of large runtime collections: ForeachParentCtx::from_state clones the entire ExecutionState (including step_results HashMap, contexts Vec, inputs HashMap, and resume_ctx), but fork_child() immediately discards all of those cloned collections by creating new empty ones. For long workflows this creates unnecessary overhead on every foreach step.
</details>
<details>

- **warning** `runkon-flow/src/executors/script.rs:191` — Subprocess stdout/stderr are piped but only read after child.wait() returns. If a script writes more data than the OS pipe buffer (typically ~64KB) to either stdout or stderr, the child will block on write and never exit. The parent then deadlocks waiting for the child to finish. This is a denial-of-service (hang/availability) regression compared to the previous temp-file approach.
</details>
<details>

- **warning** `runkon-flow/src/executors/script.rs:205` — execute_script now implements the previously-ignored ScriptNode.timeout field using wait_timeout, but the diff contains no regression test for the timeout path (script runs longer than timeout and is killed) or the wait_timeout/child.wait error paths. The only new test covers the sensitive-env-var security fix.
- **warning** `runkon-flow/src/engine.rs:980` — fork_child test resets child_runner to None in the parent fixture and therefore does not verify that a Some(Arc<dyn ChildWorkflowRunner>) value is correctly cloned into the child via as_ref().map(Arc::clone). If this mapping were accidentally removed, the test would still pass.
</details>



- [#2674 — String allocation on every step skip check](https://github.com/devinrosen/conductor-ai/issues/2674) — `runkon-flow/src/engine.rs:708` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Error Handling, Test Coverage, Architecture, DRY Abstraction, Security, Performance

### Blocking findings (8)

<details>

- **warning** `runkon-flow/src/engine.rs:774` — Duplicated StepSuccess mapping from WorkflowRunStep in restore_completed_step and fetch_child_completion_data
- **warning** `runkon-flow/src/executors/script.rs:75` — Script executor manually constructs a StepUpdate::completed literal instead of using the shared persist_completed_step helper
</details>
<details>
<summary><b>Error Handling</b> — 3 issues</summary>

- **critical** `runkon-flow/src/executors/script.rs:231` — `.expect("stdout was piped")` in non-test code can panic in production if the stdout pipe is unavailable after spawn.
- **critical** `runkon-flow/src/executors/script.rs:241` — `.expect("stderr was piped")` in non-test code can panic in production if the stderr pipe is unavailable after spawn.
- **warning** `runkon-flow/src/executors/mod.rs:15` — `p_err` wraps persistence errors as `EngineError::Persistence(e.to_string())` without operation context, making it hard to identify which DB operation failed from logs alone.
</details>
<details>
<summary><b>Test Coverage</b> — 3 issues</summary>

- **warning** `runkon-flow/src/types.rs:222` — New public method `StepSuccess::from_action_output` has no unit test in the diff.
- **critical** `runkon-flow/src/executors/script.rs:240` — Pipe-deadlock fix (spawning stdout/stderr reader threads before wait()/wait_timeout()) lacks a regression test.
- **warning** `runkon-flow/src/executors/script.rs:222` — New error paths in `execute_script` (spawn failure, wait()/wait_timeout() failure, non-UTF-8 stdout/stderr, reader thread panic) are not covered by existing tests.
</details>



- [#2675 — Manager constructors have inconsistent arity](https://github.com/devinrosen/conductor-ai/issues/2675) — `conductor-core/src/repo.rs:89` (warning)
- [#2676 — Library crate contains a binary target](https://github.com/devinrosen/conductor-ai/issues/2676) — `conductor-core/Cargo.toml:33` (warning)
- [#2677 — Web route executes raw SQL instead of using domain managers](https://github.com/devinrosen/conductor-ai/issues/2677) — `conductor-web/src/routes/stats.rs:35` (critical)
- [#2678 — Push subscription manager defined in binary crate](https://github.com/devinrosen/conductor-ai/issues/2678) — `conductor-web/src/push.rs:25` (critical)
- [#2679 — TUI presentation concern leaked into core config](https://github.com/devinrosen/conductor-ai/issues/2679) — `conductor-core/src/config.rs:350` (warning)
- [#2680 — Foreach executor also inlines StepUpdate::completed instead of shared helper](https://github.com/devinrosen/conductor-ai/issues/2680) — `runkon-flow/src/executors/foreach.rs:236` (warning)
- [#2681 — Unbounded memory growth in check_stuck](https://github.com/devinrosen/conductor-ai/issues/2681) — `runkon-flow/src/engine.rs:849` (warning)
- [#2682 — parallel.rs spawns raw OS threads instead of using a thread pool](https://github.com/devinrosen/conductor-ai/issues/2682) — `runkon-flow/src/executors/parallel.rs:216` (warning)

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Error Handling, Test Coverage, Security, DRY/Abstraction, Architecture, Performance

### Blocking findings (5)

<details>
<summary><b>DRY/Abstraction</b> — 3 issues</summary>

- **warning** `runkon-flow/src/executors/foreach.rs:236` — Manually constructs a completed StepUpdate literal instead of reusing the existing persist_completed_step helper
- **warning** `runkon-flow/src/engine.rs:131` — fork_child repeats every field of ExecutionState even though the struct now derives Clone
- **warning** `runkon-flow/src/executors/script.rs:237` — Identical pipe-reader thread spawn is duplicated for stdout and stderr
</details>
<details>

- **warning** `runkon-flow/src/types.rs:262` — New public method `StepSuccess::from_workflow_run_step` lacks a unit test in the diff.
- **warning** `runkon-flow/src/executors/script.rs:71` — New script error paths (non-success exit with stderr capture, non-UTF-8 stderr) introduced by the pipe-based refactor are not covered by tests in the diff.
</details>

devinrosen (commented):

**Reviewed by:** DB Migrations, File Types, Test Coverage, Security, Error Handling, DRY Abstraction, Architecture, Performance

### Blocking findings (6)

<details>
<summary><b>review-architecture</b> — 1 issue</summary>

- **warning** `runkon-flow/src/executors/parallel.rs:216` — parallel executor still uses std::thread::spawn while foreach was updated to use threadpool. Both dispatch concurrent work; inconsistent threading strategies leave parallel without any bound on OS thread count.
</details>
<details>
<summary><b>review-dry-abstraction</b> — 2 issues</summary>

- **warning** `runkon-flow/src/engine.rs:131` — ExecutionState::fork_child manually resets ~20 runtime fields one by one. If a new accumulator or collection field is added to ExecutionState, the author must remember to clear it here or child workflows will inherit stale parent runtime state. The existing unit test is exhaustive for current fields but cannot catch future omissions.
- **warning** `runkon-flow/src/types.rs:249` — StepSuccess::from_workflow_run_step ignores the metric fields (cost_usd, num_turns, duration_ms, input_tokens, output_tokens, cache_read_input_tokens, cache_creation_input_tokens) that are available on WorkflowRunStep. This is inconsistent with from_action_output, which copies all relevant fields. Callers such as restore_completed_step must manually extract and pass those same seven fields to accumulate_metrics before calling the helper, partially defeating the consolidation benefit.
</details>
<details>
<summary><b>review-test-coverage</b> — 3 issues</summary>

- **warning** `runkon-flow/src/engine.rs:166` — accumulate_metrics now returns bool to indicate whether any metrics were present, but no test covers the false path (all None inputs) or verifies that flush_metrics is gated correctly.
- **warning** `runkon-flow/src/executors/script.rs:201` — The refactored execute_script adds explicit error handling for cmd.spawn() failure, stdout/stderr pipe unavailability, and child.wait() failure, but none of these new error paths are exercised by the new tests.
- **warning** `runkon-flow/src/executors/foreach.rs:451` — New dependents_map logic for promoting waiting items to ready introduces a new code path (only checking direct dependents of terminal items) without test coverage.
</details>



- [#2683 — parallel.rs uses unbounded raw threads while foreach.rs uses a threadpool](https://github.com/devinrosen/conductor-ai/issues/2683) — `runkon-flow/src/executors/parallel.rs:0` (warning)
- [#2684 — N+1 DB writes in foreach fan-out item updates](https://github.com/devinrosen/conductor-ai/issues/2684) — `runkon-flow/src/executors/foreach.rs:422` (warning)

Issue #2728, opened 2026-04-30
bug: ticket_id and repo_id lost on child workflow runs spawned from foreach


devinrosen:
**Severity:** bug (data lineage)
**Location:** \`conductor-core/src/workflow/runkon_bridge.rs:425-426\` (read sites), \`runkon-flow/src/executors/foreach.rs:530-535\` (call site), \`runkon-flow/src/engine.rs:131-153\` (\`fork_child\`)

## Symptom

When a workflow runs a \`foreach\` block, every child workflow run dispatched per-item has \`ticket_id = None\` and \`repo_id = None\` on the \`workflow_runs\` row, even when the parent workflow run was associated with a ticket and/or repo. The child run loses its lineage to the parent's ticket.

The same workflow expressed as a \`call workflow\` (no \`foreach\`) preserves both fields correctly. Only the \`foreach\` dispatch path is affected.

## Root cause

\`runkon_bridge.rs::ConductorChildWorkflowRunner::execute_child\` resolves these fields from the parent state's \`inputs\` HashMap:

\`\`\`rust
// runkon_bridge.rs:425
ticket_id: parent_state.inputs.get(\"ticket_id\").cloned(),
repo_id: parent_state.inputs.get(\"repo_id\").cloned(),
\`\`\`

Two call sites in the engine pass different \`parent_state\` values:

- **\`call_workflow.rs:263\`** passes the actual parent \`ExecutionState\` — \`inputs\` is intact, lookup succeeds.
- **\`foreach.rs:530-535\`** passes a *forked* \`child_state\` (built via \`ExecutionState::fork_child\`, called transitively through \`ForeachParentCtx::make_child_state\`). \`fork_child\` does \`child.inputs.clear()\` — so the lookup returns \`None\` regardless of what the parent had.

\`\`\`rust
// engine.rs:131
pub fn fork_child(&self, ...) -> ExecutionState {
    let mut child = self.clone();
    child.inputs.clear();   // ← culprit for the foreach path
    // ...
}
\`\`\`

## Suggested fix

Pass the real parent state to \`execute_child\`. Once #2597 lands and the trait takes \`&ChildWorkflowContext\`, change \`foreach.rs\` to build the context from the *real* parent's \`state\` (with intact inputs), not from a forked \`child_state\`. The fork inside the engine is for accumulating metrics / cancellation, not for the bridge view.

[...]

PR #2729, opened 2026-04-30
refactor(runkon-flow): narrow ChildWorkflowRunner trait to ChildWorkflowContext (#2597)

### LINKED ISSUE #2597: ChildWorkflowRunner::execute_child() exposes ExecutionState through trait boundary (runkon-flow/src/engine.rs:0)
**Severity:** warning
**Location:** runkon-flow/src/engine.rs:0
**Found by:** Architecture

The `ChildWorkflowRunner` trait passes `parent_state: &ExecutionState` — the engine's full internal state struct — to `execute_child()`. Bridge implementations in conductor-core access at least six distinct fields on it. Any rename or restructuring of `ExecutionState` fields becomes a breaking change for every `ChildWorkflowRunner` implementor. Suggestion: introduce a narrower `ChildWorkflowContext { worktree_ctx, exec_config, inputs, triggered_by_hook, event_sinks }` type carrying only the subset bridge code actually needs.

devinrosen:
## Problem

The `ChildWorkflowRunner` trait was passing `&ExecutionState` — the engine's full mutable runtime state, ~30 public fields — across a public seam:

\`\`\`rust
// runkon-flow/src/engine.rs (before)
pub trait ChildWorkflowRunner: Send + Sync {
    fn execute_child(
        &self,
        workflow_name: &str,
        parent_state: &ExecutionState,    // ← engine's full internal state
        params: ChildWorkflowInput,
    ) -> Result<WorkflowResult>;
    // ...
}
\`\`\`

Conductor-core's bridge (the only real implementor today) reads **11 distinct fields**:

| Field | |
|---|---|
| `worktree_ctx.{working_dir, repo_path, worktree_id, extra_plugin_dirs}` | 4 |
| `event_sinks`, `exec_config` | 2 |
| `inputs.get("ticket_id")`, `inputs.get("repo_id")` | 2 |
| `model`, `target_label`, `triggered_by_hook`, `workflow_run_id` | 4 |

All reads, no mutations. Any rename or restructuring of `ExecutionState` fields was a breaking change for every harness.

## Change

Introduce `ChildWorkflowContext` — a narrow stable surface carrying exactly the 8-field subset the bridge actually consumes:

\`\`\`rust
#[derive(Clone)]
pub struct ChildWorkflowContext {
    pub worktree_ctx: WorktreeContext,
    pub workflow_run_id: String,
    pub model: Option<String>,
    pub target_label: Option<String>,
    pub exec_config: WorkflowExecConfig,
    pub inputs: HashMap<String, String>,
    pub triggered_by_hook: bool,
    pub event_sinks: Arc<[Arc<dyn EventSink>]>,
}

impl ExecutionState {
    pub fn child_workflow_context(&self) -> ChildWorkflowContext { /* clones */ }
}
\`\`\`

Trait signature switches from `&ExecutionState` to `&ChildWorkflowContext`. Both engine call sites (`call_workflow.rs`, `foreach.rs`) pass `&state.child_workflow_context()`. Three test mocks updated (`MockChildRunner`, `CancellingMockRunner`, `DummyChildRunner`).

### Owned vs borrowed

Owned struct with cloned fields. Per-projection cost is microseconds against multi-second child workflow runs, and the bridge already paid an `event_sinks.iter().cloned().collect()` clone. Borrowed (`<'a>`) was rejected as it would add a lifetime parameter to every harness's `execute_child` signature.

## Behavior preservation

This is a refactor only. The `foreach` call site continues to project from the *forked* `child_state` (built via `ExecutionState::fork_child`, which clears `inputs`), so foreach children still have `ticket_id = None` / `repo_id = None`. **#2728 tracks the lineage fix** — a one-line change in `foreach.rs` to project from the parent's real state instead, becomes trivial once this lands.

## Test plan

- [x] `cargo test -p runkon-flow --features test-utils` → 227 lib + 38 integration passed
- [x] `cargo test -p conductor-core --lib` → 1926 passed
- [x] `cargo clippy --all-targets --features test-utils -- -D warnings` → clean on `runkon-flow` and `conductor-core`
- [x] `cargo fmt --all --check` → clean
- [x] `cargo build` → clean for `conductor-cli`, `conductor-tui`, `runkon-runtimes`

Closes #2597. Unblocks #2728.

### DIFF
--- a/conductor-core/src/workflow/runkon_bridge.rs
+++ b/conductor-core/src/workflow/runkon_bridge.rs
@@ -453,6 +453,7 @@ impl runkon_flow::engine::ChildWorkflowRunner for ConductorChildWorkflowRunner {
         &self,
         workflow_run_id: &str,
         model: Option<&str>,
+        parent_ctx: &runkon_flow::engine::ChildWorkflowContext,
     ) -> runkon_flow::engine_error::Result<runkon_flow::types::WorkflowResult> {
         let input = crate::workflow::types::WorkflowResumeInput {
             config: &self.config,
@@ -461,7 +462,7 @@ impl runkon_flow::engine::ChildWorkflowRunner for ConductorChildWorkflowRunner {
             from_step: None,
             restart: false,
             conductor_bin_dir: None,
-            event_sinks: vec![],
+            event_sinks: parent_ctx.event_sinks.iter().cloned().collect(),
             db_path: Some(self.db_path.clone()),
             shutdown: None,
         };
--- a/conductor-core/src/workflow/runkon_bridge.rs
+++ b/conductor-core/src/workflow/runkon_bridge.rs
@@ -385,6 +385,32 @@ impl ConductorChildWorkflowRunner {
             conn,
         }
     }
+
+    /// Project a parent's `ChildWorkflowContext` into the `WorkflowResumeInput`
+    /// that `super::coordinator::resume_workflow` consumes.
+    ///
+    /// Extracted so the `event_sinks` propagation is unit-testable without
+    /// spinning up a real workflow run — see the regression test in
+    /// `tests::resume_input_propagates_event_sinks_from_parent_ctx` which
+    /// guards against `event_sinks: vec![]` re-creeping back in.
+    fn build_resume_input<'a>(
+        &'a self,
+        workflow_run_id: &'a str,
+        model: Option<&'a str>,
+        parent_ctx: &runkon_flow::engine::ChildWorkflowContext,
+    ) -> crate::workflow::types::WorkflowResumeInput<'a> {
+        crate::workflow::types::WorkflowResumeInput {
+            config: &self.config,
+            workflow_run_id,
+            model,
+            from_step: None,
+            restart: false,
+            conductor_bin_dir: None,
+            event_sinks: parent_ctx.event_sinks.iter().cloned().collect(),
+            db_path: Some(self.db_path.clone()),
+            shutdown: None,
+        }
+    }
 }
 
 impl runkon_flow::engine::ChildWorkflowRunner for ConductorChildWorkflowRunner {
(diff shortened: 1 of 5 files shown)


devinrosen:
Addressed in `f1b2de3b`:

**Finding 1 (Architecture, `resume_child` event_sinks)** — fixed. Extended `resume_child` to take `&ChildWorkflowContext` symmetric with `execute_child`. Bridge now threads `parent_ctx.event_sinks` into `WorkflowResumeInput`, and `call_workflow.rs:137` passes `&state.child_workflow_context()` at the call site. Resumed child workflows now emit step events to TUI/web consumers.

**Finding 2 (Error Handling, `load_workflow_by_name` map_err)** — pushing back. The format string is `"failed to load sub-workflow '{}': {e}"` — the `{e}` interpolation expands the inner error's `Display` impl, so the error message text is preserved end-to-end. What's "lost" is the typed error chain (`source()`), but `EngineError::Workflow(String)` is the codebase's intentional wire format throughout this file — `wrap_child_workflow_err` (lines 446, 470 in the same impl) does the same `format!("{ctx}: {other}")` on `ConductorError`. Moving to typed source-chain preservation would be a module-wide refactor and is out of scope for #2597. Happy to file a follow-up if it's worth tracking.

devinrosen:
Round 2 addressed in `f106be45`:

**Architecture — `foreach.rs:46` Arc sharing** ✅ Fixed. `make_child_state` now gives each clone a fresh `current_execution_id` Arc instead of sharing one across all parallel siblings via `template.clone()`. Pre-existing bug-in-waiting (today the bridge projects to `ChildWorkflowContext` and rebuilds state downstream, so the shared Arc never gets read), but cheap defense against future regressions.

**Test Coverage (critical) — `runkon_bridge.rs:465` resume_child event_sinks** ✅ Fixed. Extracted `ConductorChildWorkflowRunner::build_resume_input` as a private helper so the projection is unit-testable. Two new tests:
- `resume_input_propagates_event_sinks_from_parent_ctx` — guards against `event_sinks: vec![]` re-creeping in
- `resume_input_with_empty_parent_sinks_yields_empty_sinks` — positive control

**Test Coverage — `child_workflow_context()` projection** ✅ Fixed. Added `child_workflow_context_projects_all_eight_fields` asserting every projected field equals the source on the parent `ExecutionState`, and that `event_sinks` is shared via `Arc::ptr_eq`, not deep-copied.

**Test Coverage — `ChildWorkflowContext` struct unit test** ⏸️ Pushing back. The struct is plain data with no behavior — no constructor, no methods, no validation. The only thing a struct test can assert is "the fields exist and have the right types," which the projection test above already exercises end-to-end on every field. Adding a separate struct test would be duplicative.

```
cargo test -p runkon-flow --features test-utils       → 228 lib + 38 integ. ✓
cargo test -p conductor-core --lib                    → 1928 passed ✓
```

devinrosen:
Round 3 addressed in `8ace5504`:

**Test Coverage — `foreach.rs:50` Arc isolation regression test** ✅ Fixed.

Added `make_child_state_isolates_current_execution_id_per_child`:
- Builds a parent `ExecutionState`, constructs `ForeachParentCtx::from_state`, calls `make_child_state` twice
- Asserts `!Arc::ptr_eq(child1.current_execution_id, child2.current_execution_id)`
- Behavioral check: writing into child1's slot does not appear in child2's

Manually verified the test catches the precise regression: removing the fix line (`child.current_execution_id = Arc::new(...)`) causes the `Arc::ptr_eq` assertion to fire; restoring it returns to green.

```
cargo test -p runkon-flow --features test-utils       → 229 lib + 38 integ. ✓
cargo test -p conductor-core --lib                    → 1928 passed ✓
```

Issue #2732, opened 2026-04-30
design: lease-based ownership for workflow runs to eliminate watchdog/engine races


devinrosen:
**Status:** Design — implementation split into 5 sub-tickets
**Theme:** correctness / multi-worker readiness

## Sub-tickets (implementation order)

| PR | Issue | Status |
|---|---|---|
| 1/5 | #2792 — add `owner_token` / `lease_until` / `generation` columns | Ready (qualified) |
| 2/5 | #2793 — acquire lease at `FlowEngine::run` / `resume` entry, fail-fast on conflict | Blocked on 1/5 |
| 3/5 | #2794 — lease refresh thread + abort-on-refresh-failure | Blocked on 2/5 |
| 4/5 | #2795 — generation check on step writes, abort cleanly on 0-row update | Blocked on 3/5 |
| 5/5 | #2796 — watchdog claims expired leases only + subprocess kill on lease loss | Blocked on 4/5 |

This issue closes when 5/5 merges.

---

## Problem

Today's workflow run model has no explicit notion of "who currently owns this run." A row carries `status = 'running'` and `last_heartbeat`, and any process that sees stale state can grab it. The recovery query in [`WorkflowManager::claim_heartbeat_stuck_runs`](../blob/main/conductor-core/src/workflow/manager/recovery.rs) is the only gate, and it's a heuristic — "no children running + last_heartbeat older than threshold + no parent" — not a hard claim.

Consequence: any time the engine pauses between steps (or in a long wait, or during a slow DB write, or during GC, or on a slow machine) longer than the heartbeat threshold, the watchdog can claim the run while the original engine thread is still alive. Two threads then advance the same run, producing duplicate step rows, duplicate Claude agent runs (real cost), and duplicate side effects (PR review posts, GitHub comments, etc.).

#2731 documents one concrete occurrence. That fix (heartbeat refresh during parallel/foreach waits) closes the most common case but does not eliminate the race class — any new long internal pause reopens it. Tuning thresholds is racing against ourselves.

## Proposal: lease-based ownership

Standard leader-lease pattern (Cassandra, etcd, k8s coordination). Make ownership an explicit, atomic, time-bounded claim:

```sql
ALTER TABLE workflow_runs ADD COLUMN owner_token TEXT;
ALTER TABLE workflow_runs ADD COLUMN lease_until TEXT;
ALTER TABLE workflow_runs ADD COLUMN generation  INTEGER NOT NULL DEFAULT 0;
```

[...]

PR #2734, opened 2026-04-30
fix(engine): tick heartbeat during parallel/foreach waits (#2731)

### LINKED ISSUE #2731: bug: heartbeat watchdog races engine during long parallel/foreach waits, double-running steps
**Severity:** bug (correctness, cost)
**Location:** `runkon-flow/src/executors/{parallel,foreach}.rs` (wait loops), `conductor-core/src/workflow/manager/recovery.rs:652` (`claim_heartbeat_stuck_runs`)

## Symptom

Workflow runs whose parallel or foreach blocks take longer than the 60 s heartbeat threshold can be reclaimed by `claim_heartbeat_stuck_runs` while the engine thread is still alive. The reclaimed thread runs forward from "after the last completed step" in parallel with the original engine thread, producing duplicate step rows, duplicate agent runs (= duplicate Claude cost), and duplicate side effects (e.g. duplicate PR review submissions).

## Concrete observation

Workflow run `01KQDYX89400WDGH7B5704V2EV` (review-pr on PR #2729):

\`\`\`
position  step_name              status     n   notes
--------  ---------------------  ---------  --  --------------------------------------------
0         detect-db-migrations   completed  1
1         detect-file-types      completed  1
2-7       review-* (6 agents)    completed  1   parallel block, ~4 min duration
8         review-db-migrations   skipped    2   ← duplicate
9         review-aggregator      completed  2   ← duplicate (both spawned Claude — 2x cost)
10        submit-review          completed  2   ← duplicate (likely double-posted PR review)
\`\`\`

Timeline:

| Time | Event |
|---|---|
| 01:09:03.706 | Parallel block dispatches 6 reviewers + skips review-db-migrations (#1 skip) |
| 01:09:39 → 01:13:03.303 | The 6 agents finish; last is review-architecture |
| **01:13:03.639** | **review-aggregator #1 fires** — 0.34 s after last agent ended, parallel step rows still settling |
| 01:13:05.522-523 | Parallel step rows finish updating to 'completed' |
| 01:13:05.524 | review-aggregator #2 fires + db-migrations skip #2 (original engine thread) |
| 01:13:57 / 01:14:01 | Both review-aggregator agents end (2 real Claude calls) |
| 01:13:59 / 01:14:03 | Both submit-review scripts run |

Cost: ~\$0.44 wasted per occurrence (one full extra review-aggregator agent run).

## Root cause

`WorkflowManager::claim_heartbeat_stuck_runs` query:

\`\`\`sql
WHERE status = 'running'
  AND parent_workflow_run_id IS NULL
  AND NOT EXISTS ( ... wrs.status IN ('running','pending','waiting') )
  AND (now - last_heartbeat) > :threshold_secs   -- 60s
\`\`\`

When the last parallel child's row flips to `completed`, there is a window where:

- No step row is in running/pending/waiting (all 6 just settled)
- `last_heartbeat` has not been refreshed during the 4-minute parallel wait
- Run status is still 'running' (engine has not yet advanced past the block)

The watchdog claims the run and spawns `spawn_heartbeat_resume`, which picks up "after the last completed step" and runs review-aggregator + submit-review. Meanwhile the original engine thread is also advancing past the parallel block.

`last_heartbeat` is held in `Arc<AtomicI64>` (`runkon-flow/src/engine.rs:78`) but is only updated at step boundaries. Long parallel/foreach waits do not refresh it, so any wait > threshold is vulnerable.

## Fix

Bump `last_heartbeat` from inside the parallel and foreach wait loops, e.g. every 5 – 10 s while children are still running. Same liveness signal the rest of the engine emits between steps; just extends the property to "while waiting for children" too.

This is the immediate fix. **#2732** tracks the long-term structural fix that makes the race impossible by construction. The two are complementary — heartbeat refresh during waits stays the right behavior even after leases land, since lease renewals are themselves a periodic liveness signal.

## Acceptance criteria

- [ ] Long-running parallel blocks (> 60 s, > 5 min, > 30 min) do not get reclaimed by `claim_heartbeat_stuck_runs` while the engine is alive.
- [ ] Test: run a parallel block whose children sleep > threshold; assert exactly one set of post-block step rows.
- [ ] Same for foreach.

## Related

- Workflow run `01KQDYX89400WDGH7B5704V2EV` is the captured evidence.
- See companion roadmap issue for the structural fix (lease-based ownership).

devinrosen:
## Bug

The watchdog reaper (`claim_heartbeat_stuck_runs`) races the engine after >60 s when a parallel block or foreach fan-out has no in-flight steps but children are still running. The query:

\`\`\`sql
WHERE status = 'running'
  AND parent_workflow_run_id IS NULL
  AND NOT EXISTS (... wrs.status IN ('running','pending','waiting') )
  AND (now - last_heartbeat) > 60
\`\`\`

…matches whenever the last step row settled but the run hasn't finished. `last_heartbeat` only got updated at step boundaries, so a multi-minute parallel block with the children's step rows already `completed` looked stale. The reaper claimed the run, spawned a resume, and ran the rest of the workflow alongside the original engine — duplicate step rows, duplicate agent runs (= duplicate Claude cost), and duplicate side effects (e.g. PR reviews posted twice on workflow run \`01KQDYX89400WDGH7B5704V2EV\`).

## Fix

Extract the throttled heartbeat-and-cancel block from `execute_nodes` into `ExecutionState::tick_heartbeat_throttled()` and call it from both wait loops.

**foreach.rs:** existing 50 ms `recv_timeout` loop gets `let _ = state.tick_heartbeat_throttled();` at the top. The helper sets `state.cancellation` on external cancel; the existing `is_cancelled()` check at line 472 drives the controlled drain — no early-return needed.

**parallel.rs:** convert `for ... in completion_rx` → `loop { recv_timeout(500 ms) }`:
- On `Ok` → process result, mirror fail_fast as before
- On `Timeout` → tick; on external cancel, propagate to `scope_token` (same shape as fail_fast) and keep draining
- On `Disconnected` → break

Helper preserves the engine main loop's exact early-return-on-cancel semantics via `?` propagation. Throttle (5 s) and TTL (60 s) keep ample margin against the watchdog threshold even on slow machines.

## Tests

- **Helper unit tests** (`engine.rs::tests`): first call ticks, immediate second call is throttled; external cancel sets `state.cancellation` and returns `Err(Cancelled)`.
- **parallel integration test**: `SleepingExecutor` sleeps 1300 ms, asserts ≥1 tick during the wait. **Bidirectional verification** — disabling the helper call in parallel.rs causes the test to fail.
- **foreach integration test**: `SleepingChildRunner` sleeps 800 ms, asserts ≥1 tick during the wait. Same bidirectional check.
- New `CountingPersistence` helper in `test_helpers.rs` — wraps `InMemoryWorkflowPersistence` and counts `tick_heartbeat` calls.

## Test plan

- [x] `cargo test -p runkon-flow --features test-utils` → 233 lib + 38 integ. (3 new tests; 0 regressions)
- [x] `cargo test -p conductor-core --lib` → 1928 passed
- [x] `cargo clippy --all-targets -F test-utils -- -D warnings` → clean
- [x] `cargo fmt --all --check` → clean

Closes #2731. Complementary to #2732 (lease-based ownership, the long-term structural fix that closes the entire race class — heartbeat refresh stays the right thing for the engine to do regardless).

### DIFF
--- a/runkon-flow/src/engine.rs
+++ b/runkon-flow/src/engine.rs
@@ -152,6 +152,64 @@ impl ExecutionState {
         Arc::new(AtomicI64::new(0))
     }
 
+    /// Throttled heartbeat tick + external cancel check.
+    ///
+    /// Bumps `last_heartbeat` in persistence at most once every 5 seconds and
+    /// polls for cross-process cancellation via `persistence.is_run_cancelled`.
+    /// On external cancel, sets `self.cancellation` and returns
+    /// `Err(EngineError::Cancelled)`.
+    ///
+    /// Callers that own the engine main loop use `?` to propagate cancellation
+    /// up. Wait loops that need to drain in-flight work (parallel, foreach)
+    /// can call this best-effort and rely on `self.cancellation.is_cancelled()`
+    /// for their controlled exit — the cancellation token is set by this
+    /// helper before the `Err` is returned.
+    ///
+    /// Without this being called from inside long-running wait loops (parallel
+    /// blocks, foreach fan-out), the heartbeat goes stale during multi-minute
+    /// waits and the watchdog reaper races the engine after >60 s — see #2731.
+    pub fn tick_heartbeat_throttled(&self) -> Result<()> {
+        use crate::cancellation_reason::CancellationReason;
+
+        let now_secs = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_else(|e| {
+                tracing::warn!("system clock regressed: {e}; heartbeat suppressed");
+                e.duration()
+            })
+            .as_secs() as i64;
+        let last = self.last_heartbeat_at.load(Ordering::Relaxed);
+        if now_secs - last < 5 {
+            return Ok(());
+        }
+        self.last_heartbeat_at.store(now_secs, Ordering::Relaxed);
+        match self.persistence.is_run_cancelled(&self.workflow_run_id) {
+            Ok(true) => {
+                tracing::info!(
+                    "Workflow run {} cancelled externally, stopping execution",
+                    self.workflow_run_id
+                );
+                self.cancellation
+                    .cancel(CancellationReason::UserRequested(None));
+                return Err(EngineError::Cancelled(CancellationReason::UserRequested(
+                    None,
+                )));
+            }
+            Ok(false) => {}
+            Err(e) => {
+                tracing::warn!(
+                    "Database error during cancellation check for workflow run {}: {}",
+                    self.workflow_run_id,
+                    e
+                );
+            }
+        }
+        if let Err(e) = self.persistence.tick_heartbeat(&self.workflow_run_id) {
+            tracing::warn!("tick_heartbeat failed (non-fatal): {e}");
+        }
+        Ok(())
+    }
+
     /// Project this state into the narrow surface a `ChildWorkflowRunner`
     /// implementation needs to spawn a child run.
     pub fn child_workflow_context(&self) -> ChildWorkflowContext {
@@ -505,7 +563,6 @@ pub fn execute_nodes(
     nodes: &[WorkflowNode],
     respect_fail_fast: bool,
 ) -> Result<()> {
-    use crate::cancellation_reason::CancellationReason;
     for node in nodes {
         if respect_fail_fast && !state.all_succeeded && state.exec_config.fail_fast {
             break;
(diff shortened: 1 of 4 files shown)


devinrosen:
Review addressed in \`ef4d4065\`:

**Code Duplication — \`runkon-flow/src/test_helpers.rs:1\` ExecutionState boilerplate** ✅ Fixed. Extracted \`test_helpers::make_test_execution_state(persistence, run_id)\` with all fields defaulted; the three tests now override only what they need (\`action_registry\` for parallel, \`child_runner\` + \`registry\` for foreach, nothing extra for the engine helper test). Net -76 lines.

**Test Coverage — \`conductor-web/src/routes/mod.rs:28\` CORS configuration** ⏸️ Pushing back. This file is not in the PR's diff:

\`\`\`bash
$ git diff release/0.10.0..HEAD --stat
 runkon-flow/src/engine.rs             | 203 ++++++++++--
 runkon-flow/src/executors/foreach.rs  | 178 +++++++++
 runkon-flow/src/executors/parallel.rs | 163 +++++++-
 runkon-flow/src/test_helpers.rs       | 167 +++++++++
\`\`\`

The Test Coverage agent prompt at \`.conductor/agents/review-test-coverage.md\` already includes:

> **Scope constraint:** Work from the git diff only — do NOT open or read source files… Despite any other instructions, do NOT populate \`off_diff_findings\`.

The model leaked an off-diff finding anyway. Will file a follow-up to tighten the agent prompt so this doesn't recur — separate from this PR.

\`\`\`
cargo test -p runkon-flow --features test-utils       → 233 lib + 38 integ. ✓
\`\`\`

PR #2737, opened 2026-04-30
feat(workflow): inject resolved PR base branch as {{base_branch}} (#2736)

### LINKED ISSUE #2736: Inject base branch as a workflow variable so reviewer agents don't have to compute it
**Severity:** enhancement (correctness, eliminates a recurring race)
**Source:** workflow run \`01KQE8SYYNJN1WGD846E148JS6\` review-architecture leak; quick fix shipped in PR #2735.

## Problem

Reviewer agents currently detect the PR's base branch by running shell commands:

\`\`\`bash
BASE_BRANCH=\$(gh pr view --json baseRefName -q .baseRefName ...)
\`\`\`

This is brittle because the LLM agent often \`cd\`s into a different directory before running the lookup (observed: agents \`cd\`'d to \`/Users/devin/Personal/conductor-ai\` — the canonical repo path — even though cwd was already the worktree). From the wrong cwd, \`gh pr view\` finds no matching PR, the \`|| echo main\` fallback fires, and the diff gets computed against \`main\` instead of the actual base (\`release/0.10.0\`).

PR #2735 hardened the snippet (use \`gh pr list --head <branch>\` + drop silent fallback) and Layer C of submit-review.sh deterministically filters hallucinated findings against the actual diff. But the agent still wastes a full review on the wrong diff and may emit dozens of false findings before they get filtered.

## Proper fix

Pre-resolve the base branch in a workflow script step that runs once at the start of \`review-pr\`, and inject the result as a substituted variable available to all subsequent agents.

\`\`\`
script resolve-pr-base {
  run = \".conductor/scripts/resolve-pr-base.sh\"
  output = \"pr-base\"
}

parallel {
  with = [\"review-diff-scope\", \"off-diff-findings\"]
  ...
}
\`\`\`

\`prompts/review-diff-scope.md\` would then use a literal:

\`\`\`bash
git diff origin/{{base_branch}}...HEAD
\`\`\`

…with no shell needed. Deterministic — the script step runs in the worktree's correct cwd, hits gh once, and the value is frozen for all downstream agents.

## Implementation sketch

1. New script \`.conductor/scripts/resolve-pr-base.sh\`:
   \`\`\`bash
   BRANCH=\$(git rev-parse --abbrev-ref HEAD)
   BASE=\$(gh pr list --head \"\$BRANCH\" --state open --json baseRefName -q '.[0].baseRefName')
   [ -z \"\$BASE\" ] && exit 1
   echo \"<<<FLOW_OUTPUT>>>{\\\"markers\\\":[\\\"base_resolved\\\"], \\\"context\\\":\\\"\$BASE\\\", \\\"metadata\\\":{\\\"base_branch\\\":\\\"\$BASE\\\"}}<<<END_FLOW_OUTPUT>>>\"
   \`\`\`

2. Plumb the structured output into a workflow-scoped variable. The current variable injection (\`prompt_builder.rs\`) supports \`{{prior_context}}\`, \`{{prior_contexts}}\`, \`{{prior_output}}\`, etc. — adding a way for a script step to expose a *named* variable consumable by later steps would be the smallest engine change.

   Alternative: parse \`{{prior_output}}\` JSON in each consumer. Less ergonomic but no engine change.

3. Update \`prompts/review-diff-scope.md\` to use \`{{base_branch}}\` and remove the shell example entirely.

4. Same fix for \`detect-file-types.sh\` / \`detect-db-migrations.sh\` — they have the same \`|| echo main\` pattern; should consume the resolved value rather than recompute.

## Related

- PR #2735 — quick-fix predecessor (snippet hardening + script filter).
- Workflow run \`01KQE8SYYNJN1WGD846E148JS6\` — the leak that surfaced this.

## Acceptance criteria

- [ ] Agents never run \`gh pr view\` themselves to compute the base branch.
- [ ] A single script step computes the base once at the start of \`review-pr\` (and any other reviewer workflow that needs it).
- [ ] Detection failure aborts the workflow rather than silently using \`main\`.
- [ ] A test confirms reviewer agents see the correct base branch even when an agent \`cd\`s elsewhere mid-review.

devinrosen:
## Problem

Reviewer agents and detect-* scripts in `review-pr.wf` each ran `gh pr view --json baseRefName` themselves, with a silent `|| echo main` fallback. Concrete failure on workflow run \`01KQE8SYYNJN1WGD846E148JS6\`: a reviewer agent \`cd\`'d into the canonical repo path before the lookup, \`gh pr view\` returned nothing, the fallback produced a diff against \`main\`, and the reviewer then critiqued ~150 files of code accumulated on \`release/0.10.0\` as if it belonged to that PR.

PR #2735 hardened the snippet (\`gh pr list --head\` keys on explicit branch + drop silent fallback) but the fundamental issue remained: the LLM agent does the resolution.

## Fix

Resolve the base branch **once** in a script step at the start of \`review-pr.wf\`, expose it to every downstream step via the engine's variable substitution layer as \`{{base_branch}}\`. No agent or detect-* script computes it themselves; they just consume the value.

### Engine plumbing (\`runkon-flow\`)

- **\`FlowOutput\` gains \`extras: HashMap<String, serde_json::Value>\`** via \`#[serde(flatten, default)]\`. Unknown top-level fields round-trip through parse + re-serialize.
- **\`executors/script.rs\` now persists \`structured_output\`** for FLOW_OUTPUT-emitting scripts (was \`None\`). The full parsed FlowOutput is re-serialized so extras land in the step record + ContextEntry.
- **\`prompt_builder::build_variable_map\` walks prior contexts** looking for \`base_branch\` in any structured_output JSON and exposes it as \`{{base_branch}}\`. Targeted (not a generic exports mechanism) — easy to extend if more named values come up.

### Workflow + scripts

- **New \`.conductor/scripts/resolve-pr-base.sh\`**: runs \`gh pr list --head \$(git rev-parse --abbrev-ref HEAD)\`. Exits 1 with a loud message on failure — no silent \`main\` fallback. Emits FLOW_OUTPUT with \`{"base_branch": "<branch>"}\`.
- **\`review-pr.wf\`**: \`script resolve-pr-base\` is now the first step. \`detect-file-types\` and \`detect-db-migrations\` consume \`BASE_BRANCH\` via \`env = { BASE_BRANCH = "{{base_branch}}" }\`.
- **\`prompts/review-diff-scope.md\`**: replaced the shell-based detection with a literal \`git diff origin/{{base_branch}}...HEAD\`. No shell logic in the snippet at all.
- **\`detect-file-types.sh\` / \`detect-db-migrations.sh\`**: read BASE_BRANCH from env, fail loud if unset.

## Tests

- \`helpers::extras_fields_are_preserved_on_parse_and_serialize\` — round-trip.
- \`prompt_builder::build_variable_map_exposes_base_branch_from_prior_context\` — covers no-prior, present, overwrite-by-later, missing-doesn't-clobber.
- \`prompt_builder::substitute_uses_base_branch_from_variable_map\` — end-to-end template substitution.

## Smoke tests

- \`resolve-pr-base.sh\` against a known PR (\`fix/2728-foreach-child-ticket-lineage\`) → returns \`release/0.10.0\` ✓
- \`resolve-pr-base.sh\` with no PR for the branch → exits 1 with a clear message ✓
- \`cargo run --bin conductor -- workflow validate review-pr\` → PASS ✓

## Test plan

- [x] \`cargo test -p runkon-flow --features test-utils\` → 236 lib + 38 integ. (3 new tests; 0 regressions)
- [x] \`cargo test -p conductor-core --lib\` → 1928 passed
- [x] \`cargo clippy --all-targets -F test-utils -- -D warnings\` → clean
- [x] \`cargo fmt --all --check\` → clean
- [ ] Live: this PR's review-pr run will exercise the new path end-to-end.

Closes #2736.

### DIFF
--- a/.conductor/prompts/review-diff-scope.md
+++ b/.conductor/prompts/review-diff-scope.md
@@ -2,29 +2,14 @@
 
 Get the diff for this PR using the appropriate command for the review scope:
 
-- If the scope is **full** (default): detect the PR base branch from the
-  current branch's open PR and diff against it. **Run the command from the
-  current working directory — do NOT `cd` anywhere.** The worktree is
-  already cwd; `cd`ing into a different checkout (e.g. the canonical repo
-  path) silently breaks `gh pr view`, which then falls back to `main` and
-  produces a wrong diff.
+- If the scope is **full** (default): the PR's base branch has already been
+  resolved by the workflow's `resolve-pr-base` step and is injected as
+  `{{base_branch}}`. Use it literally — do not run `gh pr view` or compute
+  the base yourself.
 
   ```bash
-  # Resolve the PR's base branch. Use `gh pr list --head` rather than
-  # `gh pr view` so the lookup keys on the explicit branch name and is not
-  # affected by cwd or the currently checked-out HEAD of an unrelated repo.
-  BRANCH=$(git rev-parse --abbrev-ref HEAD)
-  BASE_BRANCH=$(gh pr list --head "$BRANCH" --state open \
-                  --json baseRefName -q '.[0].baseRefName' 2>/dev/null)
-  if [ -z "$BASE_BRANCH" ]; then
-    echo "ERROR: could not resolve PR base branch for $BRANCH — aborting" >&2
-    exit 1
-  fi
-  git diff "origin/${BASE_BRANCH}...HEAD"
+  git diff "origin/{{base_branch}}...HEAD"
   ```
-
-  Do NOT silently fall back to `main` if base detection fails. A wrong base
-  produces a fabricated diff that contaminates every downstream finding.
 
 - If the scope is **incremental**: run `git diff HEAD~1` to see only the latest commit.
 
--- a/.conductor/scripts/detect-db-migrations.sh
+++ b/.conductor/scripts/detect-db-migrations.sh
@@ -1,11 +1,16 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Get the PR's actual base branch (falls back to main for non-PR contexts)
-base_branch=$(gh pr view --json baseRefName -q '.baseRefName' 2>/dev/null || echo "main")
+# BASE_BRANCH is injected by the workflow from the resolve-pr-base step.
+# We refuse to silently fall back to a guess — a wrong base produces a
+# fabricated diff that contaminates the downstream review. See #2736.
+if [ -z "${BASE_BRANCH:-}" ]; then
+  echo "ERROR: BASE_BRANCH env var not set — workflow must run resolve-pr-base first." >&2
+  exit 1
+fi
 
 # Get changed files relative to the PR base branch
-changed_files=$(git diff "origin/${base_branch}...HEAD" --name-only 2>/dev/null || true)
+changed_files=$(git diff "origin/${BASE_BRANCH}...HEAD" --name-only 2>/dev/null || true)
 
 # Filter for migration files
 migration_files=()
--- a/.conductor/scripts/detect-file-types.sh
+++ b/.conductor/scripts/detect-file-types.sh
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Get changed files relative to PR base branch
-BASE_BRANCH=$(gh pr view --json baseRefName -q .baseRefName 2>/dev/null || echo main)
-changed_files=$(git diff origin/${BASE_BRANCH}...HEAD --name-only 2>/dev/null || true)
+# BASE_BRANCH is injected by the workflow from the resolve-pr-base step.
+# We refuse to silently fall back to a guess — a wrong base produces a
+# fabricated diff that contaminates every conditional reviewer. See #2736.
+if [ -z "${BASE_BRANCH:-}" ]; then
+  echo "ERROR: BASE_BRANCH env var not set — workflow must run resolve-pr-base first." >&2
+  exit 1
+fi
+
+changed_files=$(git diff "origin/${BASE_BRANCH}...HEAD" --name-only 2>/dev/null || true)
 
 # Filter for code files, excluding .conductor/, docs/, .github/, and root-level *.md
 code_files=()
--- a/.conductor/scripts/resolve-pr-base.sh
+++ b/.conductor/scripts/resolve-pr-base.sh
@@ -30,8 +30,11 @@ if [ -z "${BASE_BRANCH}" ]; then
   exit 1
 fi
 
-cat <<EOF
-<<<FLOW_OUTPUT>>>
-{"markers": ["base_branch_resolved"], "context": "${BASE_BRANCH}", "base_branch": "${BASE_BRANCH}"}
-<<<END_FLOW_OUTPUT>>>
-EOF
+# Build the FLOW_OUTPUT JSON via `jq -n --arg` so any quotes / backslashes /
+# control chars in BASE_BRANCH are encoded safely. String interpolation here
+# would let a crafted branch name (e.g. one containing `\` or `"`) inject
+# extra keys into the parsed FlowOutput.
+PAYLOAD=$(jq -nc --arg base "${BASE_BRANCH}" \
+  '{markers: ["base_branch_resolved"], context: $base, base_branch: $base}')
+
+printf '<<<FLOW_OUTPUT>>>\n%s\n<<<END_FLOW_OUTPUT>>>\n' "${PAYLOAD}"
--- a/.conductor/scripts/resolve-pr-base.sh
+++ b/.conductor/scripts/resolve-pr-base.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# resolve-pr-base.sh — resolve the PR's base branch ONCE per workflow run and
+# expose it to downstream steps via the engine's variable substitution layer.
+#
+# Emits a FLOW_OUTPUT block whose `base_branch` field is picked up by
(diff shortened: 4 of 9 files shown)


devinrosen:
Round 2 addressed in \`1389dc8c\`:

**Architecture — generic extras walk** ✅ Fixed.

The reviewer's point landed: \`FlowOutput.extras\` was open-ended but the consumer was hardcoded to one key, and every future script-exported variable would have meant another engine-side code change. Switched to a generic walk.

**Two coordinated changes:**

1. **\`HashMap<&str, String>\` → \`HashMap<String, String>\`** for the variable map and \`substitute_variables*\` / \`resolve_child_inputs\` signatures. The borrow model was a small premature optimization at this call frequency; \`String\` keys allow dynamic insertion from prior contexts without \`Box::leak\` or lifetime gymnastics. Ripple was contained to runkon-flow (4 files; conductor-core's separate prompt_builder uses \`<&str, &str>\` and is untouched).

2. **Generic walk in \`build_variable_map\`**: iterate every prior context's \`structured_output\` and expose each string-valued top-level field (other than \`markers\`/\`context\`) as a \`{{name}}\` variable. New scripts can plumb values just by emitting them in FLOW_OUTPUT — no engine code change.

**Shadowing guard:** keys in \`ENGINE_INJECTED_KEYS\` (\`workflow_run_id\`, \`ticket_id\`, \`repo_path\`, …) cannot be overwritten by scripts. Attempts log a \`tracing::warn!\` and are dropped. Prevents accidental or malicious hijacking of engine-controlled state.

**Two new tests:**
- \`build_variable_map_exposes_arbitrary_string_extras\` — covers exposure of arbitrary export names
- \`build_variable_map_blocks_engine_injected_key_shadowing\` — covers the guard, iterating \`ENGINE_INJECTED_KEYS\` so the test stays correct as that list evolves

\`\`\`
cargo test -p runkon-flow --features test-utils       → 238 lib + 38 integ ✓ (2 new)
cargo test -p conductor-core --lib                    → 1928 passed ✓
\`\`\`

devinrosen:
Round 3 addressed in \`fff34800\`:

**Error Handling — \`script.rs:357\` silent serialize failure** ✅ Fixed. \`.ok()\` replaced with \`.map_err(|e| tracing::warn!(...))\`. Names the step and explains the consequence (downstream \`{{name}}\` variable injection from this step's extras becomes unavailable).

**Error Handling — \`prompt_builder.rs\` silent parse failure** ✅ Fixed. Both the JSON-parse-error path and the non-object-payload path now log a \`tracing::warn!\` naming the step. structured_output is always written by \`script.rs\` via serde_json, so a parse failure here is an invariant violation — the warning surfaces a real bug rather than masking it.

**Security — \`resolve-pr-base.sh\` JSON injection** ✅ Fixed. Switched from string interpolation to \`jq -nc --arg base "\${BASE_BRANCH}"\`. Manually verified edge cases:

\`\`\`
feat/with"quote          → encoded as feat/with\"quote     → round-trips correctly
feat/with\backslash      → encoded as feat/with\\backslash → round-trips correctly
release/0.10.0           → unchanged
\`\`\`

\`\`\`
cargo test -p runkon-flow --features test-utils       → 238 lib + 38 integ ✓
cargo test -p conductor-core --lib                    → 1928 passed ✓
bash -n resolve-pr-base.sh                            → OK ✓
\`\`\`

devinrosen:
Round 4 addressed in \`13524535\`:

**Architecture — manual \`markers\`/\`context\` guard duplicates FlowOutput schema** ✅ Fixed. The reviewer's point landed cleanly. Switched from \`serde_json::from_str::<Value>(...).as_object()\` + manual skip-list to \`serde_json::from_str::<FlowOutput>(...)\` + \`flow_output.extras\` iteration. The named-field skip-list is now coupled to FlowOutput's serde schema by construction — if FlowOutput gains a new field (\`findings\`, \`severity\`, etc.), it lands in named-field land automatically rather than leaking through.

The change also collapses the previous "JSON not an object" branch: FlowOutput::deserialize already requires an object, so non-object payloads now fail at the parse step and hit the same single warning path.

**Test Coverage — \`prompt_builder.rs:455\` parse failure path** ✅ Added \`build_variable_map_skips_steps_with_invalid_structured_output\`. Three sequential steps:
1. Malformed JSON → skipped via parse-failure log
2. JSON array (non-object) → also skipped via same path (FlowOutput requires an object)
3. Valid extras → exports still flow, proving the loop continues past failures

**Test Coverage — \`prompt_builder.rs:471\` non-object check** ✅ Closed by the refactor — there is no separate non-object branch anymore. Same test above covers the merged path.

**Test Coverage — \`script.rs:217\` serialize error path** ⏸️ Pushing back on testability.

The serialize-failure path in \`script.rs\` is unreachable from production code:

- \`parse_flow_output\` uses \`serde_json::from_str\`, which rejects non-finite floats (NaN/Inf).
- \`serde_json::Number::from_f64(NaN)\` returns \`None\` — you cannot construct a \`Value::Number(NaN)\` through the public API.
- All other \`Value\` variants serialize unconditionally.

So a \`FlowOutput\` obtained via \`parse_flow_output\` is always re-serializable. A test would have to construct a \`FlowOutput\` by hand and stuff in something serde-rejects — which doesn't exercise the production path; it only tests \`serde_json::to_string\`. The defensive log is good practice (catches future regressions if e.g. someone adds a field with a fallible \`Serialize\` impl) but the production failure mode is empty today.

Happy to add the test if you'd prefer it for forward-coverage, but flagging the cost/benefit before doing so.

\`\`\`
cargo test -p runkon-flow --features test-utils       → 239 lib + 38 integ ✓ (1 new)
cargo test -p conductor-core --lib                    → 1928 passed ✓
\`\`\`

Issue #2633, opened 2026-04-26
conductor-tui binary crate uses rusqlite::params! directly (conductor-tui/src/app/workflow_management.rs:989)


devinrosen:
**Severity:** warning
**Location:** conductor-tui/src/app/workflow_management.rs:989
**Found by:** Architecture

conductor-tui/src/app/workflow_management.rs:989 uses rusqlite::params! in the TUI binary. Like the web layer, the TUI should not reach into rusqlite's macro surface — all query logic should live in conductor-core managers and be exposed through higher-level APIs.

devinrosen:
## ✅ Ready for Implementation

Ticket #2633 is READY. The single rusqlite::params! call at conductor-tui/src/app/workflow_management.rs:989 is confirmed present, all referenced APIs exist (WorktreeManager::get_by_id at manager.rs:568, RepoManager::get_by_id at repo.rs:162, Worktree struct has repo_id/slug/ticket_id), and the fix path is unambiguous: chain the two existing manager calls to replace the raw SQL JOIN, eliminating the layer violation.

devinrosen:
Completed via PR #2773 (merged into `release/0.11.0`, so auto-close didn't fire — closing manually).

Issue #2632, opened 2026-04-26
conductor-web binary crate depends on rusqlite directly (conductor-web/src/routes/workflows.rs:100)


devinrosen:
**Severity:** warning
**Location:** conductor-web/src/routes/workflows.rs:100
**Found by:** Architecture

notify_workflow (line 100) and a second helper (line 2837) in conductor-web/src/routes/workflows.rs both accept &rusqlite::Connection as a parameter. Binary crates should be thin consumers of conductor-core managers and must not import rusqlite directly. These functions should be moved into conductor-core behind a manager or notification-service API, exposing only higher-level types to the web layer.

devinrosen:
## ❓ Open Questions

The following questions or issues must be resolved before this ticket can be handed off to an autonomous agent:

Ticket 2632 is NOT READY. Codebase scan confirms two real production violations (notify_workflow in workflows.rs:99 and open_db_and_config in worktrees.rs:20), but the ticket misclassifies insert_agent_run (line 2836) and seed_conversations (line 386) as violations — both are test helpers inside #[cfg(test)]. The ticket is blocked by unresolved architecture decisions: how conductor-web should obtain database connections without importing rusqlite directly (re-export from conductor-core, opaque wrapper, or path-based API), whether test helpers are in scope, and whether rusqlite should be fully removed from Cargo.toml. These choices produce substantially different code and require a human decision before an agent can proceed.

devinrosen:
**Decision: move the two `&Connection`-accepting functions into conductor-core and expose them through a manager or service**

The two functions in `conductor-web/src/routes/workflows.rs` that take `&rusqlite::Connection` directly should move to conductor-core — likely as methods on `WorkflowManager` or `WorkflowQueryService` (per #2590 split plan).

Steps:
1. Identify the two functions (lines ~100 and ~2837)
2. Move them into the appropriate conductor-core module
3. Update conductor-web call sites to use the core API
4. Verify `rusqlite` can be removed from conductor-web's `Cargo.toml` (coordinate with #2678 — both contribute to that dep)

devinrosen:
**Revised decision: delete `notify_workflow` wrapper and move `rusqlite` to dev-dependencies**

The issue cited two functions but on inspection only one is a production violation:

- `notify_workflow` (line 100) — real production violation: a thin wrapper that accepts `&rusqlite::Connection` solely to call `fire_workflow_notification`, which already lives in `conductor-core/src/notify/runs.rs`
[...]

PR #2773, opened 2026-05-01
refactor(tui): replace raw SQL JOIN with managed get_by_id_enriched (closes #2627, #2633)

### LINKED ISSUE #2627: TUI binary executes raw SQL against conductor-core DB schema (pre-existing layer violation) (conductor-tui/src/app/workflow_management.rs:984)
**Severity:** warning
**Location:** conductor-tui/src/app/workflow_management.rs:984
**Found by:** Architecture

Lines 984–989 of `workflow_management.rs` execute a raw SQL JOIN against `worktrees` and `repos` tables directly inside the TUI binary, bypassing the manager layer entirely. `WorktreeManager::get_by_id_enriched(&worktree_id)` already returns `repo_slug`, `wt_slug`, and `ticket_id` via the managed layer and should be used instead. Raw SQL in a binary crate against conductor-core's internal schema is a clear layer violation — schema changes in core will silently break this query at runtime rather than at compile time.

devinrosen:
## Summary

The cache-miss fallback in `do_dispatch_workflow` at `conductor-tui/src/app/workflow_management.rs:971` ran a raw SQL JOIN against `worktrees ⋈ repos` directly inside the TUI binary, bypassing the manager layer entirely and using `rusqlite::params!` against conductor-core's internal schema.

Replaces the inline query with `WorktreeManager::get_by_id_enriched` + `RepoManager::get_by_id`. Same behavior, no raw SQL, no `rusqlite::params!` in this binary crate. Future schema changes in core will now fail at compile time rather than at runtime.

## Closes

- #2627 — TUI binary executes raw SQL against conductor-core DB schema
- #2633 — `rusqlite::params!` used directly in TUI binary

These two issues describe different facets of the **same** 6-line block, so they're addressed together.

## Diff

| | |
|---|---:|
| Files changed | 1 |
| Insertions | +13 |
| Deletions | -12 |
| Net | **+1 line** |

## Test plan

- [x] `cargo build -p conductor-tui --lib` — passes
- [x] `cargo clippy -p conductor-tui -- -D warnings` — clean
- [x] `cargo test -p conductor-tui --lib` — 195 tests pass, 0 failures
- [x] `cargo test -p conductor-core --lib` — 1946 tests pass, 0 failures (sanity, no behavior change in core)

## Notes

- Surviving `rusqlite` mentions in `conductor-tui/src/`: `app/mod.rs:7` and `background.rs:24` both use `rusqlite::Connection` as a type — that's the manager-pattern threading concern (a binary crate must hold a `Connection` to call any manager), not a layer violation in the same sense. Out of scope for this PR.
- `notify_workflow` in `conductor-web` and friends (#2632) are a separate cluster covered by another PR.

### DIFF
--- a/conductor-tui/src/app/workflow_management.rs
+++ b/conductor-tui/src/app/workflow_management.rs
@@ -969,22 +969,23 @@ impl App {
                             .map(|r| (Some(format!("{}/{}", r.slug, w.slug)), w.ticket_id.clone()))
                     })
                     .unwrap_or_else(|| {
-                        // Cache miss — query DB directly to avoid storing "".
-                        use conductor_core::config::db_path;
+                        // Cache miss — go through the managed layer rather than raw SQL.
+                        use conductor_core::config::{db_path, load_config};
                         use conductor_core::db::open_database;
+                        use conductor_core::repo::RepoManager;
                         let label = (|| -> Option<WorktreeLabelParts> {
                             let conn = open_database(&db_path()).ok()?;
-                            let (repo_slug, wt_slug, ticket_id): (String, String, Option<String>) =
-                                conn.query_row(
-                                    "SELECT r.slug, w.slug, w.ticket_id \
-                                     FROM worktrees w \
-                                     JOIN repos r ON r.id = w.repo_id \
-                                     WHERE w.id = ?1",
-                                    rusqlite::params![worktree_id],
-                                    |row| Ok((row.get(0)?, row.get(1)?, row.get(2)?)),
-                                )
+                            let config = load_config().ok()?;
+                            let wt = WorktreeManager::new(&conn, &config)
+                                .get_by_id_enriched(&worktree_id)
                                 .ok()?;
-                            Some((Some(format!("{repo_slug}/{wt_slug}")), ticket_id))
+                            let repo = RepoManager::new(&conn, &config)
+                                .get_by_id(&wt.worktree.repo_id)
+                                .ok()?;
+                            Some((
+                                Some(format!("{}/{}", repo.slug, wt.worktree.slug)),
+                                wt.worktree.ticket_id,
+                            ))
                         })();
                         label.unwrap_or((None, None))
                     });
--- a/conductor-tui/src/app/workflow_management.rs
+++ b/conductor-tui/src/app/workflow_management.rs
@@ -969,17 +969,14 @@ impl App {
                             .map(|r| (Some(format!("{}/{}", r.slug, w.slug)), w.ticket_id.clone()))
                     })
                     .unwrap_or_else(|| {
-                        // Cache miss — go through the managed layer rather than raw SQL.
-                        use conductor_core::config::{db_path, load_config};
-                        use conductor_core::db::open_database;
+                        // Cache miss — go through the managed layer rather than raw SQL,
+                        // reusing the App's existing `conn` and `config` rather than reopening.
                         use conductor_core::repo::RepoManager;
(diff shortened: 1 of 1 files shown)


devinrosen (commented):
Triage of the review-swarm findings:

**Architecture (`get_by_id_enriched` over-fetches): accepted.** Switched to plain `WorktreeManager::get_by_id` in 0184570. The enriched variant pulls 23 columns and runs an `INNER JOIN ... MAX(started_at) GROUP BY` agent-runs subquery + a `LEFT JOIN tickets` — none of which this closure reads. `get_by_id` returns just the worktree row.

**Performance (over-fetch / N+1): accepted, fixed by the same change.** "N+1" is mislabeled here (no loop, just two queries instead of one), but the substantive concern — over-fetching plus the unneeded agent-runs subquery — goes away with `get_by_id`. The remaining "two queries vs one JOIN" only runs on cache miss (just-after-create state), so the cost is negligible.

**Error Handling (silent `.ok()?`): declined for this PR; tracked in #2774.** This is pre-existing behavior — the original raw-SQL version also swallowed errors via `.ok()?`. Changing it here would expand a refactor PR's scope into error-semantics. Filed #2774 as a focused follow-up to add a single boundary `tracing::warn!` so failures become observable without changing the "best-effort, don't crash dispatch" contract.

devinrosen:
Triage of the review-swarm findings:

**Architecture (`get_by_id_enriched` over-fetches): accepted.** Switched to plain `WorktreeManager::get_by_id` in 01845705. The enriched variant pulls 23 columns and runs an `INNER JOIN ... MAX(started_at) GROUP BY` agent-runs subquery + a `LEFT JOIN tickets` — none of which this closure reads. `get_by_id` returns just the worktree row.

**Performance (over-fetch / N+1): accepted, fixed by the same change.** "N+1" is mislabeled here (no loop, just two queries instead of one), but the substantive concern — over-fetching plus the unneeded agent-runs subquery — goes away with `get_by_id`. The remaining "two queries vs one JOIN" only runs on cache miss (just-after-create state), so the cost is negligible.

**Error Handling (silent `.ok()?`): declined for this PR; tracked in #2774.** This is pre-existing behavior — the original raw-SQL version also swallowed errors via `.ok()?`. Changing it here would expand a refactor PR's scope into error-semantics. Filed #2774 as a focused follow-up to add a single boundary `tracing::warn!` so failures become observable without changing the "best-effort, don't crash dispatch" contract.

devinrosen:
Triage of the second review-swarm pass:

**Architecture (reuse `App::conn` / `App::config`): accepted.** Fixed in e0979a89. The closure now captures `&self.conn` and `&self.config` directly — disjoint-field borrowing makes this safe even though `&self.state` is held by the outer iterator chain. Removed the `open_database` + `load_config` calls entirely.

**Error Handling (silent `load_config()`): resolved as a side-effect of the architecture fix.** With `self.config` reused, there's no `load_config()` call to fail. The two remaining `.ok()?` calls (`get_by_id` × 2) are still silent — that's the same pre-existing concern already tracked in #2774 (which I've updated to reflect the smaller surface).

**Performance (config I/O + 2 queries): the I/O concern is resolved.** Cache miss no longer opens a DB connection or reads config from disk. Two `get_by_id` calls remain — that's still cache-miss-only (just-after-create state) and trivially cheap on the existing connection. Calling this "N+1" remains a misnomer (no loop), and at this point the path is about as light as it gets without merging the two lookups into a single core API method, which would be over-engineering for a rarely-hit fallback.

Diff for the second pass: +4 / -7.

PR #2835, opened 2026-05-03
feat(workflow): per-run engine log + panic hook (#2821)

devinrosen:
## Summary

- **Per-run log file** (`~/.conductor/workflow-logs/<run_id>.log`): each engine thread installs a scoped `tracing` subscriber via `with_default`, writing structured logs to a per-run file without touching the host binary's global subscriber or any other engine thread
- **Panic capture**: `catch_unwind` wraps the engine closure in `spawn_workflow_resume` and `spawn_heartbeat_resume`; panics are logged to the per-run file and persisted to `workflow_runs.error` (status flipped to `failed`) before being re-raised so the host panic handler still fires

The body-position progression signal (item 3 in #2821) was split off as a follow-up — see the rollback commit. The signal as designed had no consumer (reaper, watchdog, observability), and the boundaries it captured were already marked by step rows + heartbeats. It will be re-litigated paired with a concrete consumer and defined action in a separate ticket.

## Test plan

- [x] `cargo test --workspace` — green (conductor-core 1980/1980, runkon-flow 263/263, conductor-tui 195/195, conductor-web 126/126, tui_snapshots 21/21)
- [x] `cargo test -p conductor-core -- workflow::engine_log` — two unit tests: `engine_panic_is_recorded` (verifies log file + DB row flip) and `log_file_created_on_normal_run`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean

Closes #2821 (item 3 split into follow-up)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -670,6 +670,8 @@ dependencies = [
  "tokio",
  "toml 0.8.23",
  "tracing",
+ "tracing-appender",
+ "tracing-subscriber",
  "tracing-test",
  "ulid",
  "ureq",
@@ -5420,6 +5422,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "symlink"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7973cce6668464ea31f176d85b13c7ab3bba2cb3b77a2ed26abd7801688010a"
+
+[[package]]
 name = "syn"
 version = "1.0.109"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6168,6 +6176,19 @@ dependencies = [
  "pin-project-lite",
  "tracing-attributes",
  "tracing-core",
+]
+
+[[package]]
+name = "tracing-appender"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "050686193eb999b4bb3bc2acfa891a13da00f79734704c4b8b4ef1a10b368a3c"
+dependencies = [
+ "crossbeam-channel",
+ "symlink",
+ "thiserror 2.0.18",
+ "time",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/conductor-core/Cargo.toml
+++ b/conductor-core/Cargo.toml
@@ -22,6 +22,8 @@ ureq = { version = "2", features = ["json"] }
 serde_yml = "0.0"
 notify = "6"
 tracing = "0.1"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+tracing-appender = "0.2"
 web-push = "0.10"
 libc = "0.2"
 tempfile = "3"
--- a/conductor-core/src/config.rs
+++ b/conductor-core/src/config.rs
@@ -526,6 +526,23 @@ pub fn agent_log_path(run_id: &str) -> Result<PathBuf> {
     Ok(agent_log_dir().join(format!("{run_id}.log")))
 }
 
+/// Returns the directory for per-run workflow engine log files.
+pub fn workflow_log_dir() -> PathBuf {
+    conductor_dir().join("workflow-logs")
+}
+
+/// Returns the log file path for a given workflow run ID.
+///
+/// Convention: `~/.conductor/workflow-logs/{run_id}.log`
+///
+/// Returns an error if `run_id` is not a valid ULID, preventing path traversal.
+pub fn workflow_log_path(run_id: &str) -> Result<PathBuf> {
+    run_id
+        .parse::<ulid::Ulid>()
+        .map_err(|_| ConductorError::Workflow(format!("invalid run_id: {run_id}")))?;
+    Ok(workflow_log_dir().join(format!("{run_id}.log")))
+}
+
 impl Config {
     /// Returns the Anthropic API key from the `ANTHROPIC_API_KEY` environment variable.
     ///
--- a/conductor-core/src/config.rs
+++ b/conductor-core/src/config.rs
@@ -1748,6 +1748,20 @@ bot_name = "my-bot"
         assert!(err.to_string().contains("invalid run_id"));
     }
 
+    #[test]
+    fn workflow_log_path_valid_ulid_returns_ok() {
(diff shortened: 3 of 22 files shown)


devinrosen:
Pushed 485a3ac5 addressing the second review. Five of seven blocking findings are resolved in code; two are responded to here.

**Resolved in code:**
- **Arch #3** — \`record_panic_in_db\` now sets \`ended_at\` alongside \`status\`/\`error\`. Updated test asserts \`ended_at IS NOT NULL\` after a panic.
- **Error #1** — Distinguishes \`Ok(0)\` from \`Ok(_)\`: when the \`status='running'\` guard matches no rows (another path got there first), we emit a \`tracing::warn!\` so the gap is auditable. Status guard kept on purpose so we don't overwrite a deliberate \`cancelled\`/\`failed\` with a late panic. New test \`already_terminal_run_does_not_overwrite_status\` pins this behavior.
- **Security #1 + Test #1** — \`setup_per_run_tracing\` now routes through \`config::workflow_log_path\` for ULID validation; on invalid input we warn and fall back to a sanitized basename. Resolves the dead-public-API concern from prior reviews. Added two tests for \`workflow_log_path\` (valid/invalid ULID).
- **Test #2** — Three new error-path tests: \`missing_db_does_not_propagate_panic_handler_error\`, \`unwritable_log_dir_does_not_panic\`, \`already_terminal_run_does_not_overwrite_status\`.

**Responded to (not changed):**

- **Arch #2 (\`GlobalForwardLayer\` only forwards events)** — Documented as a deliberate design choice. Cross-Subscriber span-ID forwarding would create broken span trees on the host side: span IDs are scoped to the Subscriber that allocated them, so forwarding our local \`new_span\`/\`enter\`/\`exit\` calls to the host's Dispatch would arrive without a matching \`new_span\` and either panic the host's Registry or pollute its span store. The honest fix is either (a) document the events-only limitation or (b) restructure to a per-thread Layer attached to the host's global Subscriber, which conflicts with keeping engine-spawn infrastructure inside conductor-core. Engine code in this repo emits ad-hoc events, not heavy span trees, so events-only forwarding covers the common case. Doc comment on the layer makes this explicit.

- **Arch #1 (\`tracing-subscriber\`/\`tracing-appender\` in library crate)** — Defensible. The "library shouldn't depend on subscribers" guideline applies cleanly when a library is consumed by unrelated apps and shouldn't dictate subscriber strategy. \`conductor-core\` is a domain library consumed by exactly three first-party binaries in the same workspace, and the engine-thread spawn site lives in \`conductor-core/src/workflow/coordinator.rs\` (per the TUI threading rule in CLAUDE.md). The per-run subscriber install is engine-thread infrastructure — it must run at the spawn point. Splitting it across binaries would either duplicate subscriber-construction code at three boundaries or force exposing a low-level Layer API across the binary/library line. Keeping it where the engine is spawned is the right factoring.

**Off-diff #2839 — closing as invalid:** \`spawn_claimed_runs\` (line 885) delegates per-ID to \`spawn_workflow_resume\`, which IS wrapped at line 817 with \`run_engine_with_diagnostics\`. The reviewer pattern-matched on the spawn-site list in the issue body without reading the implementation. Each claimed run gets per-run log + panic capture via the wrap on \`spawn_workflow_resume\`.

Workspace test suite green (1985 conductor-core, 278 runkon-flow, 195 conductor-tui, 126 conductor-web, 21 tui_snapshots), clippy \`-D warnings\` clean, fmt clean.

devinrosen:
Filed #2840 for the layered abstraction work — moving `engine_log.rs` into runkon-flow and emitting panics as `EngineEvent::Panicked` for harnesses to handle via `EventSink`. Out of scope for this PR (release branch); both pieces fall out together cleanly when picked up.

devinrosen:
Both findings push back; the off-diff is a recurring false positive.

**Arch #1 — \`tracing-subscriber\`/\`tracing-appender\` in conductor-core.** This is the same finding from the previous review round. Same architectural argument applies: \`conductor-core\` is consumed by exactly three first-party binaries in the same workspace, and the engine-thread spawn site lives in \`conductor-core/src/workflow/coordinator.rs\` per the TUI threading rule. The per-run subscriber install is engine-thread infrastructure that runs at the spawn point.

**However**, the proper architectural fix is in flight: #2840 proposes moving \`engine_log.rs\` into \`runkon-flow\` alongside emitting panics as \`EngineEvent::Panicked\` for harnesses to consume via \`EventSink\`. That refactor moves these deps to runkon-flow where they belong — the engine itself is in runkon-flow — and removes them from \`conductor-core\` entirely. Out of scope for this release-branch PR; tracked there for the next opportunity.

**DRY #1 — \`workflow_log_path\` duplicates \`agent_log_path\`.** Borderline. The two functions are 4 lines each with 3 lines of overlap:

\`\`\`rust
pub fn agent_log_path(run_id: &str) -> Result<PathBuf> {
    run_id.parse::<ulid::Ulid>()
        .map_err(|_| ConductorError::Agent(format!("invalid run_id: {run_id}")))?;
    Ok(agent_log_dir().join(format!("{run_id}.log")))
}

pub fn workflow_log_path(run_id: &str) -> Result<PathBuf> {
    run_id.parse::<ulid::Ulid>()
        .map_err(|_| ConductorError::Workflow(format!("invalid run_id: {run_id}")))?;
    Ok(workflow_log_dir().join(format!("{run_id}.log")))
}
\`\`\`

They differ on (a) the base directory and (b) the error variant. Extracting a helper that takes a closure for the error variant would dedupe the parsing but add closure-passing overhead to two functions that are already trivial. CLAUDE.md's project guidance is explicit on this: \"Three similar lines is better than a premature abstraction.\" Two callers, both short, both self-contained, error variants intentionally differ — this is when to leave it. If a third such function ever needs the same validation, extract then.

**Off-diff #2841 — closing as duplicate of #2839.** Same false positive I closed yesterday. The auto-reviewer has now re-filed this finding twice; the source ticket #2821 listed \`spawn_claimed_runs\` in its location list as one of three engine-spawn sites, and the reviewer pattern-matches on that list without reading the implementation. \`spawn_claimed_runs\` is a fan-out helper that delegates per-ID to \`spawn_workflow_resume\` (which IS wrapped) — no work to do.

No code changes in this round.

PR #2838, opened 2026-05-03
refactor(#2679): move theme/themes_dir out of conductor-core into conductor-tui

### LINKED ISSUE #2679: TUI presentation concern leaked into core config (conductor-core/src/config.rs:350)
**Severity:** warning
**Location:** conductor-core/src/config.rs:350
**Found by:** Architecture

`GeneralConfig` contains a `theme: Option<String>` field described as "TUI color theme", plus a `themes_dir()` helper. A core library crate should not contain presentation-layer configuration (color themes, UI styles, etc.). These belong in `conductor-tui` or a shared config crate that is allowed to couple to presentation concerns.

---

## Scope (decided 2026-05-02)

Surface area is intentionally small — single PR, no sub-tickets.

### What moves
- `theme: Option<String>` field on `GeneralConfig` (conductor-core/src/config.rs:313)
- `themes_dir()` helper (conductor-core/src/config.rs:513)
- The `create_dir_all(themes_dir())` line in `ensure_dirs()` (conductor-core/src/config.rs:673)

### Where it goes
- A new `TuiConfig` struct defined in `conductor-tui` with the `theme` field.
- A `[tui]` section in the **same** `~/.conductor/config.toml` (no second config file).
- `conductor-tui` owns load/save of just its section. The existing patch-write merge logic in `save_config` already preserves unknown sections (verified by `test_save_config_preserves_unknown_sections`), so non-TUI binaries calling `save_config` will not erase the `[tui]` block.
- `themes_dir()` moves to `conductor-tui` (it is purely a TUI path helper). The TUI ensures the directory at startup; `conductor-core::ensure_dirs()` no longer creates it.

### Migration
Follow the existing deprecation-warning pattern in `load_config_from`:
- `notifications.slack.webhook_url` warning at config.rs:562
- `[notifications.workflows]` warning at config.rs:573

When `[general].theme` is present in the on-disk TOML, log a `tracing::warn!` directing users to move it to `[tui].theme`. The orphan field is silently ignored on the core side. No automatic migration writes — the field will simply be re-saved under `[tui]` the next time the user changes themes via the picker.

### Callers to update
- `conductor-tui/src/main.rs:41` — read theme name from `tui.theme` instead of `general.theme`
- `conductor-tui/src/app/theme_management.rs:53,109` — read/write through the new path
- `conductor-tui/src/theme.rs:60,291` — call the relocated `themes_dir()`

### Out of scope
- `WebPushConfig` is the same kind of leak (purely `conductor-web`, never read by core). Tracked separately so each move stays a clean, single-PR change.
- `auto_start_agent`, `agent_permission_mode`, `model`, `inject_startup_context`, `claude_config_dir`, `auto_resume_limit`, `stale_workflow_minutes`, `auto_cleanup_merged_branches` all stay on `GeneralConfig` — they drive agent runtime / workflow / sync behavior used by all binaries, not presentation.

### Acceptance criteria
- `conductor-core` has no references to `theme` or `themes_dir`
- `ensure_dirs()` does not create `~/.conductor/themes/`
- Existing configs with `[general].theme` load successfully with a deprecation warning; user re-applying a theme via the TUI picker writes it under `[tui]`
- TUI snapshot tests pass; theme picker saves and reloads correctly
- `conductor-cli` and `conductor-web` build and run unchanged

devinrosen:
## Summary

- Removes `theme: Option<String>` from `GeneralConfig` and `themes_dir()` from `conductor-core` — these are TUI presentation concerns that have no business in the core library consumed by `conductor-cli` and `conductor-web`
- `ensure_dirs()` in core no longer creates `~/.conductor/themes/`
- Adds a `tracing::warn!` deprecation in `load_config_from` when `[general].theme` is present on disk (mirrors the existing slack/notifications.workflows deprecation pattern); no auto-migration write
- New `conductor-tui/src/config.rs` owns the `[tui]` section of `~/.conductor/config.toml`: `TuiConfig { theme }`, `themes_dir()`, `ensure_tui_dirs()`, `load_tui_config()` (with legacy `[general].theme` fallback so existing users keep their theme), `save_tui_config()` (toml::Value patch-write that preserves all other sections)
- Adds `tui_config: TuiConfig` field to `App`; updates `App::new` and all construction sites
- `theme_management.rs` and `settings_management.rs` read/write through `self.tui_config`; `theme.rs` calls `crate::config::themes_dir()`

## Test plan

- [ ] `cargo nextest run -p conductor-core --features test-helpers` — all existing config tests pass; new deprecation warn fires on `[general].theme` presence
- [ ] `cargo nextest run -p conductor-tui` — 746 tests pass including 6 new `config::tests::*` tests covering load/save round-trip, legacy fallback, `[tui]` precedence over `[general]`, and unknown-section preservation
- [ ] `cargo build -p conductor-cli` and `cargo build -p conductor-web` (without frontend) compile cleanly — neither references `theme` or `themes_dir`
- [ ] No `theme` or `themes_dir` references remain in `conductor-core`

Closes #2679

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -708,6 +708,7 @@ dependencies = [
  "ratatui",
  "runkon-runtimes",
  "rusqlite",
+ "serde",
  "serde_json",
  "serde_yml",
  "tempfile",
--- a/conductor-core/src/config.rs
+++ b/conductor-core/src/config.rs
@@ -308,10 +308,6 @@ pub struct GeneralConfig {
     /// into agent prompts. Defaults to true; set to false to disable.
     #[serde(default = "default_true")]
     pub inject_startup_context: bool,
-    /// TUI color theme. One of: "conductor" (default), "nord", "gruvbox", "catppuccin_mocha",
-    /// or the stem of a file in `~/.conductor/themes/`. Omit to use the default conductor theme.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub theme: Option<String>,
     /// Automatically detect merged PRs and clean up worktrees (delete local/remote branch,
     /// remove worktree directory, auto-close orphaned features). Defaults to true.
     #[serde(default = "default_true")]
@@ -382,7 +378,6 @@ impl Default for GeneralConfig {
             agent_permission_mode: AgentPermissionMode::default(),
             model: None,
             inject_startup_context: true,
-            theme: None,
             auto_cleanup_merged_branches: true,
             stale_workflow_minutes: default_stale_workflow_minutes(),
             claude_config_dir: None,
@@ -509,11 +504,6 @@ pub fn hooks_dir() -> PathBuf {
     conductor_dir().join("hooks")
 }
 
-/// Returns the directory for user-supplied theme files: ~/.conductor/themes/
-pub fn themes_dir() -> PathBuf {
-    conductor_dir().join("themes")
-}
-
 /// Returns the log file path for a given agent run ID.
 ///
 /// Convention: `~/.conductor/agent-logs/{run_id}.log`
@@ -574,6 +564,13 @@ fn load_config_from(path: &std::path::Path) -> Result<Config> {
         tracing::warn!(
             "[notifications.workflows] is deprecated — migrate to [[notify.hooks]] with `on` patterns; \
              see CHANGELOG.md"
+        );
+    }
+
+    // Deprecation: warn if [general].theme is still present — it moved to [tui].theme.
+    if raw.get("general").and_then(|g| g.get("theme")).is_some() {
+        tracing::warn!(
+            "[general].theme is deprecated — move to [tui].theme; conductor-core no longer reads it"
         );
     }
 
@@ -670,7 +667,6 @@ fn save_config_to(config: &Config, path: &std::path::Path) -> Result<()> {
 pub fn ensure_dirs(config: &Config) -> Result<()> {
     std::fs::create_dir_all(conductor_dir())?;
     std::fs::create_dir_all(&config.general.workspace_root)?;
-    std::fs::create_dir_all(themes_dir())?;
     Ok(())
 }
 
--- a/conductor-tui/Cargo.toml
+++ b/conductor-tui/Cargo.toml
@@ -26,6 +26,7 @@ rusqlite = { version = "0.32", features = ["bundled"] }
 serde_json = "1"
 tui-textarea = "0.7"
 toml = "0.8"
+serde = { version = "1", features = ["derive"] }
 serde_yml = "0.0"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
--- a/conductor-tui/src/app/action_dispatch.rs
+++ b/conductor-tui/src/app/action_dispatch.rs
@@ -1579,6 +1579,7 @@ mod tests {
         App::new(
             conn,
             conductor_core::config::Config::default(),
+            crate::config::TuiConfig::default(),
             crate::theme::Theme::default(),
         )
     }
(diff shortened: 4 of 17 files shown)


devinrosen:
Pushed 085f7694 — one fix, one pushback.

**Addressed: Error #1 (silent \`[tui]\` deserialization)**

The \`.unwrap_or_default()\` was masking real user mistakes (e.g. \`theme = 42\` instead of a string). Replaced with explicit error handling that emits \`tracing::warn!(path, error)\` and falls back. TUI startup behavior is unchanged (still resilient — won't brick over a typo), but the misconfiguration is now auditable.

Two new tests pin the behavior:
- \`test_load_malformed_tui_section_falls_back_to_default\` — type-mismatched \`[tui]\` doesn't crash \`load_from\`.
- \`test_load_malformed_tui_falls_back_to_legacy\` — when \`[tui]\` is malformed but \`[general].theme\` is valid, the legacy fallback still recovers (regression guard for the warn-then-default path).

**Pushback: Error #2 (\`let _ = bg_tx.send(...)\`)**

This is the established TUI background-thread idiom in this repo. Quick scan of \`conductor-tui/src/background.rs\` shows 10+ instances of \`let _ = tx.send(action)\` from background threads:
- ticket sync (\`background.rs:933, 937, 944, 948, 967, 985\`)
- PR fetch / refresh (\`background.rs:1289, 1319, 1352, 1361\`)
- and more

The reasoning, per the threading rule in CLAUDE.md: a background thread sends its result Action back to the main loop on completion. \`Sender::send\` only fails if the receiver has been dropped — and the receiver lives in the App's main loop, which only drops when the TUI is exiting. So the failure mode is \"TUI is exiting,\" in which case the reviewer's framing of \"leaving UI in bad state (progress modal stuck)\" doesn't hold: there's no UI left to be stuck. Adding a \`tracing::warn!\` to every send-failure site would generate noise on every clean TUI shutdown without surfacing actionable information.

Treating \`theme_management.rs:120\` differently than the other 10+ identical patterns would be inconsistent. If we want to revisit this site-wide, that's a separate refactor — not blocking on it for this PR.

CI status: \`cargo test -p conductor-tui --lib config::\` 9/9 pass; clippy \`-D warnings\` clean; fmt clean.

devinrosen:
Pushed 70af60c8 — one fix, one pushback.

**Addressed: Error #1 (file write lacks path context)**

Every \`fs::\` and \`toml::\` call in \`load_from\` / \`save_to\` is now wrapped with \`anyhow::Context\` so failures point at the offending file. Previously \"Permission denied (os error 13)\" gave no clue which path failed; now you get \`write /Users/.../config.toml: Permission denied (os error 13)\` and similar for read/parse/create_dir paths.

**Pushback: Error #2 (\`eprintln\` + \`std::process::exit\` at \`main.rs:49\`)**

That block is **pre-existing on \`release/0.12.0\`** — this PR's diff at that location only swaps the theme source:

\`\`\`diff
-    let theme = match config.general.theme.as_deref() {
+    let theme = match tui_config.theme.as_deref() {
         Some(name) => Theme::from_name(name).unwrap_or_else(|e| {
             eprintln!(\"{e}\");
             std::process::exit(1);
\`\`\`

The error-handling pattern itself is unchanged from before this refactor; it should be tagged \`conductor-off-diff\` per the bot's own convention. Not blocking on this PR. If worth fixing, file as a separate cleanup ticket — converting it to \`?\` propagation is straightforward but it's a separate change from \"move theme out of conductor-core.\"

CI: \`cargo test -p conductor-tui --lib config::\` 9/9 pass; clippy \`-D warnings\` clean; fmt clean.

devinrosen:
Pushed 982da614 — added path context to the two remaining serialize sites (`toml::Value::try_from` and `toml::to_string_pretty`). These are in-memory serialization errors where path isn't actually the cause, but adding it for consistency with the rest of `save_to` is one line each. CI green, fmt clean.

